Issue #5, opened 2018-07-10
Suport DevServerBuilder


lucasteles:
Would be cool if you provide a wrapper of "@angular-devkit/build-angular:dev-server", for run a custom webpack on ng serve

meltedspark:
I will look into that. Meanwhile as a workaround you can use `ng build --watch` instead of `ng serve`. This is slower but does the work.

httpete:
this was exactly what I was looking for - I thought that because we replaced the browser build target ng serve would pick that up - doesn't seem to.

meltedspark:
Implemented in `2.1.0`

PR #6, opened 2018-07-11
Add dev server builder

lucasteles:
I have added DevServerBuild, but not working yet, beacause the DevServerBuilder is not using the class instance of `buildWebpackConfig`, which is fixed in this commit (https://github.com/angular/angular-cli/commit/cb588b68b34641cbfc31b563f741f94f48211b2d#diff-6c0b312b8bd33db747e50549a66f7ab1)

To this work, we have to wait they release a version with the commit code


meltedspark:
Closes #5

src/custom-webpack/dev-server/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class CustomWebpackDevServerBuilder extends DevServerBuilder {
+
+  constructor(public context: BuilderContext) {
+    super(context);
+  }
+  buildWebpackConfig(root: Path,
+                     projectRoot: Path,
+                     host: virtualFs.Host<fs.Stats>,
+                     options: NormalizedBrowserBuilderSchema) {
```
meltedspark: Should be DevServerBuilderOptions instead of NormalizedBrowserBuilderSchema

src/custom-webpack/dev-server/schema.d.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Created by Evgeny Barabanov on 01/07/2018.
+ */
+
+import {BuildWebpackServerSchema} from '@angular-devkit/build-angular/src/server/schema';
+
+export interface CustomWebpackDevServerBuildSchema extends BuildWebpackServerSchema {
```
meltedspark: Should extend DevServerBuilderOptions

src/custom-webpack/dev-server/schema.ext.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "BuildCustomWebpackServerSchema",
+  "title": "Custom webpack dev server schema for Build Facade",
+  "description": "Server target options",
```
meltedspark: Please fix the description

src/custom-webpack/dev-server/index.ts
lucasteles: Shoud be `BrowserBuilderSchema` right?

The original file is it
https://github.com/angular/angular-cli/blob/44086c60ff1d6c26d30850bef125120f6c498ac1/packages/angular_devkit/build_angular/src/dev-server/index.ts#L178

src/custom-webpack/dev-server/schema.d.ts
lucasteles: Same here, should be `BrowserBuilderSchema`

src/custom-webpack/dev-server/index.ts
meltedspark: Yep you're correct

meltedspark:
Ok, I looked into that more thoroughly and the issue is a bit deeper than we thought. Just adding a new filed to the _dev-server_ builder schema, like it is done in _custom-webpack-browser_, would be simply wrong.
_dev-server_ appears to be a generic builder that should [take build configuration](https://github.com/angular/angular-cli/blob/839c1f8a8d9a0fbe8127aecefba669bab99fc0cf/packages/angular_devkit/build_angular/src/dev-server/index.ts#L411) from the browser target you specified.  
However now it suffers from implementation issues (it is not as generic as it is supposed to be).  
Instead of returning just `browserConfig.options` and creating `BrowserBuilder` (hardcoded), they should be returning the whole browser config, getting the right builder from there (using [`architect.getBuilder`](https://github.com/angular/angular-cli/blob/839c1f8a8d9a0fbe8127aecefba669bab99fc0cf/packages/angular_devkit/architect/src/architect.ts#L337)) and getting webpack configuration from this builder.  
There are some caveats here, because builder interface has only one method `run` and doesn't have `buildWebpackConfig` method (so you have to assume that the builder from browser target does have this method). But it is all solvable. 

Later today I will open an issue in angular_devkit, but it will probably take some time until they merge the fix.  

Meanwhile we can implement this logic here (by overriding some methods from dev-server builder).  
The right name for this builder would be probably `generic-dev-server` builder and not `custom-webpack-dev-server`.

meltedspark:
Implemented in `2.1.0`

Issue #7, opened 2018-07-16
provide an option to replace a plugin 


lmessinger:
currently its only possible to add a plugin after the ng cli plugin ran. we need an option to replace a plugin rather then add one?
(eg AngularCompilerPlugin, NamedChunks, etc - for example, to allow custom, code-based naming)

meltedspark:
`custom-webpack-browser` and `custom-webpack-server` use `webpack-merge` under the hood (just as Angular CLI does). However, it uses default merge strategy, which is `append`.   
The most naive implementation if your requirement would provide the option "strategy" in custom-webpack builder configuration and then use `merge.strategy` method of `webpack-merge` for merging the configurations. One of the supported strategies is `replace`, which meets your requirement. The downside of this approach is that it would be "all or nothing", meaning the strategy would be applied to the whole configuration and this is not very flexible.  
Next step is allowing merging several configs, each with its own strategy (instead of having a single "webpackConfigPath" entry, have an array of `{webpackConfigEntry: '', strategy: ''}` entries).  
This way you'll be able to replace the part you'd like to replace and append the parts you wish to append.  
If you want you can go ahead and try implementing it yourself by creating a pull request or, alternatively, you can wait for me - I'll be available for this activity in a week.

lmessinger:
thanks very much! appreciate it. issue is - still am not able to run the default, non-replacing custom-webpack-browser

the path i went thru:
1. npm installing the pre-requisities
@angular-devkit/architect = 0.7.0-rc.0
@angular-devkit/build-angular = 0.7.0-rc.0
2. npm installing angular-cli-builders
then changing angular.json as said -
```
   "build1": {
              "builder": "angular-cli-builders:custom-webpack-browser",
              "options": {
                     "webpackConfigPath": "./extra-webpack.config.js",
                     "outputPath": "dist/my-cool-client",
                     "index": "src/index.html",
                     "main": "src/main.ts",
                     "polyfills": "src/polyfills.ts",
                     "tsConfig": "src/tsconfig.app.json"
              }},
```
but:
The specified command ("build1") is invalid.

I also tried to (1) install locally and globally (2) npm uninstall and re-install @angular/cli (3) creating a new project (ng new project) (4) npm installing globally all of node_modules\@angular-devkit packages (architect, core etc) to 0.7.0-rc.0   (5) manually replace the local project node_modules\@angular-devkit with 0.7.0-rc.0  from global 

but the issue remains

any ideas?
thanks!!

[...]

PR #8, opened 2018-07-18
Added option for merge strategies

DetectiveQuack:
Pull request for issue #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ Options:
  - All the `@angular-devkit/build-angular:browser` options
  - `webpackConfigPath`: path to the extra webpack configuration, defaults to `webpack.config.js`
  - `mergeStrategy`: webpack config merge strategy, strategies can be `append | prepend | replace`. Defaults to append
+      - `append`: adds the configurations to the end of the angular cli config. The custom plugins config will be added to the end of the angular cli plugins array
+      - `prepend`: adds the configurations to the beginning of the angular cli config. The custom loaders config will be added to the beginning of the angular cli loaders array
+      - `replace`: replaces the configuration entirely. The custom webpack config will replace the angular cli webpack config
+      See [webpack-merge](https://github.com/survivejs/webpack-merge) for more info
 
 `angular.json` Example: 
 ```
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -49,6 +49,7 @@ The builder will run the same build as `@angular-devkit/build-angular:browser` d
 Options:
  - All the `@angular-devkit/build-angular:browser` options
  - `webpackConfigPath`: path to the extra webpack configuration, defaults to `webpack.config.js`
+ - `mergeStrategy`: webpack config merge strategy, strategies can be `append | prepend | replace`. Defaults to append
```
meltedspark: Need to explain more about strategies (what does it mean and how each one works).

README.md
```diff
@@ -58,6 +59,7 @@ Options:
               "builder": "angular-cli-builders:custom-webpack-browser"
               "options": {
                      "webpackConfigPath": "./extra-webpack.config.js",
+                     "mergeStrategy": { "plugins": "prepend" },
```
meltedspark: Would be better to have replace as example. Also would be nice to explain what happens in this particular example

README.md
```diff
@@ -75,6 +77,7 @@ configuration.
 Options:
  - All the `@angular-devkit/build-angular:server` options
  - `webpackConfigPath`: path to the extra webpack configuration, defaults to `webpack.config.js`
+ - `mergeStrategy`: webpack config merge strategy, strategies can be `append | prepend | replace`. Defaults to append
```
meltedspark: Same as above

README.md
```diff
@@ -84,6 +87,7 @@ Options:
               "builder": "angular-cli-builders:custom-webpack-server"
               "options": {
                     "webpackConfigPath": "./extra-webpack.config.js",
+                    "mergeStrategy": { "plugins": "prepend" },
```
meltedspark: Same as above but can be prepend (to describe different behavior). It would be also nice to give here another example rather than plugins (for example loaders).

meltedspark:
Great work, thank you for your contribution!  
Could you please address the review comments?  
Also, it would be nice if you have added a unit test to check the merging behavior, meaning the resulting webpack config (nice to have, not necessary).

DetectiveQuack:
@meltedspark I am not sure how to write the tests other than to compare two objects and check the output after a merge.

Do you know of any other way of doing it?

PR #10, opened 2018-07-24
Added generic dev server 

scott-phillips-sp:
This PR is based upon the feedback on https://github.com/meltedspark/angular-cli-builders/pull/6. It creates a new generic dev server where if the build target is a browserBuilder then it uses that builder's public buildWebpackConfig() method to generate the webpack config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,9 @@
   "builders": "builders.json",
   "dependencies": {},
   "devDependencies": {
-    "@angular-devkit/architect": "^0.7.0-rc.0",
-    "@angular-devkit/build-angular": "^0.7.0-rc.0",
+    "@angular-devkit/architect": "^0.7.0-rc.3",
+    "@angular-devkit/build-angular": "^0.7.0-rc.3",
+    "@angular-devkit/core": "^0.7.0-rc.3",
     "@types/jest": "^23.1.4",
     "@types/lodash": "^4.14.110",
     "@types/node": "^10.5.1",
@@ -55,7 +56,8 @@
     "typescript": "^2.9.2"
   },
   "peerDependencies": {
-    "@angular-devkit/architect": ">=0.7.0-rc.0",
-    "@angular-devkit/build-angular": ">=0.7.0-rc.0"
+    "@angular-devkit/architect": ">=0.7.0-rc.3",
+    "@angular-devkit/build-angular": ">=0.7.0-rc.3",
+    "@angular-devkit/core": "^0.7.0-rc.3"
   }
 }
--- a/src/generic/dev-server/index.ts
+++ b/src/generic/dev-server/index.ts
@@ -7,8 +7,6 @@ import {tap, switchMap} from 'rxjs/operators';
 import * as fs from 'fs';
 
 
-const webpackMerge = require('webpack-merge');
-
 export class GenericDevServerBuilder extends DevServerBuilder {
 
   private targetBuilder: Builder<BrowserBuilderSchema> | undefined;
--- a/src/generic/dev-server/index.ts
+++ b/src/generic/dev-server/index.ts
@@ -4,7 +4,8 @@ import {DevServerBuilder, DevServerBuilderOptions, BrowserBuilderSchema, Browser
 import {Path, virtualFs} from '@angular-devkit/core';
 import {Observable} from 'rxjs';
 import {tap, switchMap} from 'rxjs/operators';
-import * as fs from 'fs';
+import {Stats} from 'fs';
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -55,7 +56,8 @@
     "typescript": "^2.9.2"
   },
   "peerDependencies": {
-    "@angular-devkit/architect": ">=0.7.0-rc.0",
-    "@angular-devkit/build-angular": ">=0.7.0-rc.0"
+    "@angular-devkit/architect": ">=0.7.0-rc.3",
```
meltedspark: It's preferable to have the least possible version as a dependency. Are you confident `rc.3` is the minimal required version?

src/generic/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
+
+import {BuilderContext, BuilderConfiguration, BuildEvent, Builder} from '@angular-devkit/architect';
+import {DevServerBuilder, DevServerBuilderOptions, BrowserBuilderSchema, BrowserBuilder, NormalizedBrowserBuilderSchema } from '@angular-devkit/build-angular';
+import {Path, virtualFs} from '@angular-devkit/core';
+import {Observable} from 'rxjs';
+import {tap, switchMap} from 'rxjs/operators';
+import * as fs from 'fs';
+
+
+const webpackMerge = require('webpack-merge');
```
meltedspark: Any use of this?

src/generic/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
+
+import {BuilderContext, BuilderConfiguration, BuildEvent, Builder} from '@angular-devkit/architect';
+import {DevServerBuilder, DevServerBuilderOptions, BrowserBuilderSchema, BrowserBuilder, NormalizedBrowserBuilderSchema } from '@angular-devkit/build-angular';
+import {Path, virtualFs} from '@angular-devkit/core';
+import {Observable} from 'rxjs';
+import {tap, switchMap} from 'rxjs/operators';
+import * as fs from 'fs';
```
meltedspark: Please remove unused imports

src/generic/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  buildWebpackConfig(
+    root: Path,
+    projectRoot: Path,
+    host: virtualFs.Host<fs.Stats>,
+    browserOptions: BrowserBuilderSchema,
+  ) {
+    // If our target builder has a webpack config method lets use it. Otherwise we
+    // will fall to the default config
+    if (this.targetBuilder instanceof BrowserBuilder) {
```
meltedspark: Actually I would check that the `targetBuilder` has a method `buildWebpackConfig`. It won't be necessarily instance of `BrowserBuilder`. If you go with this assumption it means everyone that wants to implement custom webpack builder will have to inherit from `BrowserBuilder`

src/generic/dev-server/schema.d.ts
```diff
@@ -0,0 +1,3 @@
+import { BrowserBuilderSchema } from '@angular-devkit/build-angular';
+
+export interface GenericDevServerBuildSchema extends BrowserBuilderSchema { }
```
meltedspark: This seems to be redundant

meltedspark:
Closes #5

package.json
scott-phillips-sp: Yes. It needs this commit, which was introduced with rc.3:

https://github.com/angular/angular-cli/commit/cb588b68b34641cbfc31b563f741f94f48211b2d#diff-6c0b312b8bd33db747e50549a66f7ab1

src/generic/dev-server/index.ts
scott-phillips-sp: 👍

mwanago:
This is just what I've needed. Great! Can't wait for it.

src/generic/dev-server/index.ts
scott-phillips-sp: fs.stat is used as part of the method signature for the buildWebpackConfig method.... Is there a better way to to import just the stat method from the fs module?

src/generic/dev-server/index.ts
scott-phillips-sp: 👍

src/generic/dev-server/index.ts
meltedspark: Yep, like this:
`import {Stats} from 'fs';` and then in sig: ` host: virtualFs.Host<Stats>,`

scott-phillips-sp:
@meltedspark Updates requested have been made. And yes I've tested this with a cli app.

PR #11, opened 2018-07-25
Add an example app that demonstrates the project

scott-phillips-sp:
https://github.com/scott-phillips-sp/angular-cli-builders/tree/example-app


meltedspark:
@scott-phillips-sp This would be great to have this app not just as an example app but also as a part of e2e  testing. Just like it is in [Jest builder](https://github.com/meltedspark/angular-builders/blob/master/packages/jest/scripts/ci.sh).   
The project has moved to a mono-repo style so there are some adjustments to do to this pull request.  
Would you like to do it yourself or I can go ahead and do that?

PR #97, opened 2018-10-08
fix: add karma/index.ts to typescript compilation

### LINKED ISSUE #96: custom-webpack:karma javascript not in package
**Describe the bug**
When using the "@angular-builders/custom-webpack:karma" builder in `angular.json`. I get the error:
`Cannot find module '<root-dir>/node_modules/@angular-builders/custom-webpack/src/karma'`.
Which is correct, because the installed dependency does have a `karma` folder, but it only has a `schema.json`, but not an `index.js`.

**To Reproduce**
It is easily reproduced by creating a new angular project and using the "@angular-builders/custom-webpack:karma" as test builder in the `angular.json`.

**Expected behavior**
I would expect the file to be included in the package.

**Builder:**
 - @angular-builders/custom-webpack:karma
 - 2.5.3

**Libraries**
 - `angular-devkit/build-angular`: 0.9.0-rc.0

njirem:
closes #96 

A possible fix for the package not containing `karma/index.js`. 

To add the file to the barrel, I had to deduplicate the `NormalizedCustomWebpackBrowserBuildSchema` interface in `karma/index.ts` and `browser/index.ts`. 
This interface is exported from it's new location, to make sure the external interface is the same as before. This may not be necessary.

### DIFF
--- a/packages/custom-webpack/src/custom-webpack-schema.ts
+++ b/packages/custom-webpack/src/custom-webpack-schema.ts
@@ -1,9 +1,9 @@
 import {CustomWebpackBuilderConfig} from "./custom-webpack-builder-config";
-import {NormalizedBrowserBuilderSchema} from "@angular-devkit/build-angular";
-
 
 export interface CustomWebpackSchema {
 	customWebpackConfig: CustomWebpackBuilderConfig;
 }
 
-export interface NormalizedCustomWebpackBrowserBuildSchema extends NormalizedBrowserBuilderSchema, CustomWebpackSchema {}
+
+
+
--- a/packages/custom-webpack/src/custom-webpack-schema.ts
+++ b/packages/custom-webpack/src/custom-webpack-schema.ts
@@ -1,9 +1,9 @@
 import {CustomWebpackBuilderConfig} from "./custom-webpack-builder-config";
+import {NormalizedBrowserBuilderSchema} from "@angular-devkit/build-angular";
+
 
 export interface CustomWebpackSchema {
 	customWebpackConfig: CustomWebpackBuilderConfig;
 }
 
-
-
-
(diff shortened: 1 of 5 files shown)


meltedspark:
Thanks for the contribution! I indeed forgot to import Karma builder in index so this pull request should fix the issue.  
Exporting `NormalizedCustomWebpackBrowserBuildSchema` from general index though will create a circular dependency (index -> karma -> index) which I prefer to avoid.  
It's completely fine to import it from browser module.

meltedspark:
Thanks for the contribution! I indeed forgot to add Karma builder to the index, so this pull request should fix the issue.

packages/custom-webpack/src/custom-webpack-schema.ts
```diff
@@ -1,9 +1,9 @@
...
+
 
 export interface CustomWebpackSchema {
 	customWebpackConfig: CustomWebpackBuilderConfig;
 }
 
-
-
-
+export interface NormalizedCustomWebpackBrowserBuildSchema extends NormalizedBrowserBuilderSchema, CustomWebpackSchema {}
```
meltedspark: Placing it here breaks the abstraction (CustomWebpackSchema is common to every module in custom webpack but BrowserSchema is not).
Probably the best place to place it is in browser/schema.d.ts and then import it from there (it's fine that Karma module depends on Browser module)

packages/custom-webpack/src/index.ts
```diff
@@ -1,5 +1,7 @@
 /**
  * Created by Evgeny Barabanov on 01/07/2018.
  */
+export {NormalizedCustomWebpackBrowserBuildSchema} from './custom-webpack-schema';
```
meltedspark: This export is not used

Issue #135, opened 2018-11-03
Export function in webpack configuration


toverux:
Migrating an Angular project with full custom webpack configuration, I'd like to use a function in my webpack configuration. When a function is passed, Webpack executes it and passes an "env" object and the argv: https://webpack.js.org/configuration/configuration-types/#exporting-a-function

I tried this:

```js
module.exports = function (env) {
    throw env;
};
```

But nothing happens, meaning that angular-builders doesn't execute it (exporting an object works, as expected).

My final goal however, was to see if I can get some info from Angular about the build, like : is the build in production mode?

I'd like to know what environment Angular uses so I can inherit from it and apply different plugins or loaders based on the environment we're in, like in the past.

Do you see a way to achieve that?

In the worst case scenario, I can use NODE_ENV and the --prod switch, but that won't match the setting Angular uses (or will it?).

meltedspark:
That is an interesting question.
Let's split it into two different features:
1. Support webpack config provided with a function rather than an object. This really shouldn't be a problem, it should be supported and it will be. 
2. Provide an access to Angular CLI execution environment within a webpack configuration. This one is not straight forward, because the webpack configuration you provide is simply merged with a webpack configuration provided by Angular (which is already configured according to the environment). Maybe the Angular environment config is already present in `process.env` (in this case it wouldn't be a problem to access it from within the webpack config) but maybe it is not. In this case it should be somehow passed in.  
I'll investigate this one and will see what can be done here.

toverux:
Thank you!
For now I simply use NODE_ENV to pilot my webpack script, en take care that --prod is set accordingly.
Meaning that I have a npm script that looks like `cross-env NODE_ENV=production ng build --prod`.
Before that, my build script was capable of doing static builds in dev or prod more, whereas now it has to be hardcoded in the command line.
So I can live without it if that's not possible, but sure that would be nice to have.

PR #152, opened 2018-11-26
feat(custom-webpack): Export function in webpack configuration

jloscos:
Add the possibility to export a function in webpack custom configuration.
The function takes the angular build option as parameter.

The need for this feature was mentioned in the issue : #135 

exemple :
````javascript
module.exports = (buildOptions) => ({
    devtool: buildOptions.sourceMap ? 'source-map' : false
});
````

### DIFF
--- a/packages/custom-webpack/src/browser/index.ts
+++ b/packages/custom-webpack/src/browser/index.ts
@@ -24,7 +24,7 @@ export class CustomWebpackBrowserBuilder extends BrowserBuilder {
                      host: virtualFs.Host<fs.Stats>,
                      options: NormalizedCustomWebpackBrowserBuildSchema): Configuration {
 	  const browserWebpackConfig = super.buildWebpackConfig(root, projectRoot, host, options);
-	  return CustomWebpackBuilder.buildWebpackConfig(root, options.customWebpackConfig, browserWebpackConfig);
+	  return CustomWebpackBuilder.buildWebpackConfig(root, options.customWebpackConfig, browserWebpackConfig, options);
   }
 }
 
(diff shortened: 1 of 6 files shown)


meltedspark:
Thanks for the contribution!  
Would you mind adding UTs that cover the functionality you've added (the function/object thing)?

jithinchandranj:
Hi,

Any update on this? This feature is required if I want to pass any run-time environment variables and bundle the resources according to my run-time parameter.

meltedspark:
@jithinchandranj This is waiting for UTs coverage and documentation to be added. 
Unfortunately currently I'm not available for this (fairly small) task, but @michaelcm kindly offered to implement it so I believe it won't take long.  
Meanwhile I suggest you to use the `configuration` option when building your project. This allows you to define several build configurations and specify each one of them in the respective section in _angular.json_ file.  
For example you can see that by default there is a `production` configuration that defines additional settings to the prod build.  
You can use the same approach to define multiple webpack configs and use them in different configurations, while calling the build with the specific configuration would depend on your environment.  
Example: `ng build --configuration=my-cool-config`

jithinchandranj:
@meltedspark That's fine. But, my scenario was to inject a custom variable that's passed as an ng-build parameter onto the webpack config.

For eg., previously I was using the build command as npm run build  --env.param=value,
and was referring it in the webpack config as:
module.exports (env) => {
};

meltedspark:
@jithinchandranj You can still do that with the existing implementation.  
[Here](https://github.com/meltedspark/electron-angular-native) is an example (look at _package.json_ in order to see how the variables are passed and at _native-webpack.config.js_ to see how they are consumed).  
If you have any further questions I'll gladly answer, don't hesitate.

jithinchandranj:
@meltedspark  Thanks buddy. I've tried it out in the same way as mentioned here. But when I run the command
ng build --param=val, compiler throws this error: Unknown option: param. 
Tried with --env.param too, but also throws the same error.

jithinchandranj:
@meltedspark Thanks mate. Got the idea.

jithinchandranj:
Hi @meltedspark ,
Another thing I wanted to ask was about the usage of webpack in multi-compiler mode. I wanted to create two separate distributions, so will have to return an array of objects from the webpack configuration. Is this supported in the library? I was not able to do so.

meltedspark:
Closed by #258

Issue #251, opened 2019-02-28
webpack.config.js not found even if customWebpackConfig is not set


chaoyangnz:
**Describe the bug**
I just want to merge extra webpack config for production build. So I only config the production options, but for development build, it runs into errors: `webpack.config.js not found`

**Builder:**
 - @angular-builders/custom-webpack
 - 7.3.1

**Libraries**
 - `angular-devkit/build-angular`: 0.12.3

**Expect**
Ignore when customWebpackConfig is not set

**Additional context**
My angular.json config is like
```
"build": {
          "builder": "@angular-builders/custom-webpack:browser",
          "options": {
            "outputPath": "dist/mall/browser",
            "index": "projects/mall/src/index.html",
            "main": "projects/mall/src/main.ts",
            "polyfills": "projects/mall/src/polyfills.ts",
            "tsConfig": "projects/mall/tsconfig.app.json",
            "sourceMap": true,
            "assets": [
              "projects/mall/src/favicon.ico",
              "projects/mall/src/assets"
            ],
            "styles": [
              "projects/mall/src/styles.scss"
            ],
            "scripts": [],
            "statsJson": true
          },
          "configurations": {
            "production": {
              "customWebpackConfig": {
                "path": "extra-webpack.config.js",
                "mergeStrategies": { "devtool": "replace" }
              }
         }
```

meltedspark:
Yeah, indeed a use case that I missed. You're welcome to PR (it's just adding a default value in schema).  
Meanwhile the workaround is to create an empty config for development and specify it in options.

meltedspark:
@chaoyangnz Hey, I saw that you closed the pull request. Does that mean that you're off this task and I should take it myself? Just checking.

chaoyangnz:
yes, please do

chaoyangnz:
The issue is not fixed. I upgrade the version, the same error can be seen.

I checked your fix, the problem is when no extraWebpackConfig is specified, its value is not `undefined` or `null`, but ```
{
    path: undefined,
    mergeStrategies: {};
    replaceDuplicatePlugins: undefined;
}
```
that is why null guard is not sufficient.

Hope this helps.

meltedspark:
Are you certain? Because there is an e2e test that passes (with unspecified value). Make sure the version you have is `7.4.2` and check out `sanity-app` in the examples.

chaoyangnz:
[...]

PR #256, opened 2019-03-01
feat(custom-webpack): #251 add a default value for customWebpackConfig and ignore if …

chaoyangnz:
for #251

### DIFF
--- a/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
+++ b/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
@@ -7,3 +7,4 @@ export const customWebpackConfig = {
 		"replaceDuplicatePlugins": {"type": "boolean", "description": "Flag that indicates whether to replace duplicate webpack plugins or not"}
 	}
 };
+
--- a/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
+++ b/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
@@ -5,7 +5,5 @@ export const customWebpackConfig = {
 		"path": {"type": "string", "description": "Path to the custom webpack configuration file"},
 		"mergeStrategies": {"type": "object", "description": "Merge strategies per webpack config field"},
 		"replaceDuplicatePlugins": {"type": "boolean", "description": "Flag that indicates whether to replace duplicate webpack plugins or not"}
-	},
-	// @ts-ignore
-	"default": null
+	}
 };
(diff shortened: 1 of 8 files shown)


packages/custom-webpack/src/schema.ext.json
```diff
@@ -7,7 +7,8 @@
         "path": {"type": "string", "description": "Path to the custom webpack configuration file"},
         "mergeStrategies": {"type": "object", "description": "Merge strategies per webpack config field"},
         "replaceDuplicatePlugins": {"type": "boolean", "description": "Flag that indicates whether to replace duplicate webpack plugins or not"}
-      }
+      },
+      "default": null
```
meltedspark: I'd rather make it `"default": {}`, this way the integration is flawless and doesn't require any additional change nor unit tests.

packages/custom-webpack/src/custom-webpack-builder.ts
```diff
@@ -7,8 +7,11 @@ export const defaultWebpackConfigPath = 'webpack.config.js';
 
 export class CustomWebpackBuilder {
 	static buildWebpackConfig(root: Path, config: CustomWebpackBuilderConfig, baseWebpackConfig: Configuration): Configuration{
+		if(!config) {
```
meltedspark: Not required if the default value is {}

packages/custom-webpack/src/custom-webpack-builder.spec.ts
```diff
@@ -55,4 +59,10 @@ describe('CustomWebpackBuilder test', () => {
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {replaceDuplicatePlugins: true}, baseWebpackConfig);
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, undefined, true);
 	});
-});
\ No newline at end of file
+
+	it('Should ignore if no customWebpackConfig set', () => {
```
meltedspark: Not required if the default value is {}

packages/custom-webpack/src/custom-webpack-builder.ts
chaoyangnz: you mean make default value {} and that's it?
but it still find `defaultWebpackConfigPath`, which is what I don't hope

packages/custom-webpack/src/schema.ext.json
meltedspark: Specifying default = null is redundant (it will be null anyways if not specified). 
So it's not necessary with the current implementation.

packages/custom-webpack/e2e/custom-webpack-config-schema.ts
```diff
@@ -5,5 +5,7 @@ export const customWebpackConfig = {
 		"path": {"type": "string", "description": "Path to the custom webpack configuration file"},
 		"mergeStrategies": {"type": "object", "description": "Merge strategies per webpack config field"},
 		"replaceDuplicatePlugins": {"type": "boolean", "description": "Flag that indicates whether to replace duplicate webpack plugins or not"}
-	}
-};
\ No newline at end of file
+	},
+	// @ts-ignore
+	"default": null
```
meltedspark: Not necessary

Issue #257, opened 2019-03-01
Modify rule regex


mistic100:
I need to change the loader used for SVG files.

Currently Angular CLI declares a rule on `\.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)$` and uses the file-loader.

**Is there a solution to alter the Regex of a an existing rule with angular-builders ?**

Previously I used ngx-build-plus and was able to alter the merge configuration with the "plugin" feature.
The plugin look likes :

```js
exports.default = {
    config: (cfg) => {
        const rules = cfg.module.rules;

        // Remove SVG from file-loader
        const fileLoader = rules.find(rule => {
            return rule.loader === 'file-loader';
        });

        if (!fileLoader) {
            throw 'Cannot find assets rule';
        }

        fileLoader.test = new RegExp(fileLoader.test.source.replace('|svg', ''), fileLoader.test.flags);

        // Add SVG to raw-loader
        rules.push({
            test  : /\.svg$/,
            loader: 'raw-loader',
        });

        return cfg;
    }
};
```

If this is not currently possible with angular-builders would you accept a Pull request to add such feature ?

meltedspark:
I'm pretty sure it is not possible now without replacing the whole `rules` array.  
Allowing custom modifications is certainly a must have feature, and I think there is already a [pull request](https://github.com/meltedspark/angular-builders/pull/152) that does something very similar.  
The only difference is that it takes build options as a parameter instead of getting webpack config.  
Making it accept both as parameters `(cfg, buildOptions)` would be the right way to go in my opinion.  

This pull request, however, lacks UT tests, E2E tests and documentation.
I'd appreciate any contribution, but it would be really great if you contributed to this pull request instead of creating a new one. What do you think?

mistic100:
Ok I can start form this PR, but it will be a new one anyway, I can't wait to know if jloscos allows me to access his repo.

meltedspark:
Huh that's right, I'm just used to having access to all the PRs 😁.
Anyway, you got my point, I think that is a common functionality which should be implemented as one piece.

mistic100:
Actually #152 is only about allowing the custom webpack file to export a function.

If I want to be able to alter the final configuration, I need a second file which will be called after "webpack-merge"

mistic100:
What do you think about a single file with this structure ?

```js
module.exports = function(buildOptions) {
   return "extra configuration";
};

module.exports.postMerge = function(config) {
[...]

PR #258, opened 2019-03-01
feat(custom-webpack): Export function in webpack configuration

### LINKED ISSUE #135: Export function in webpack configuration

mistic100:
This is a continuation of #152 and fixes #135 and #257 

See README for example usage.

### DIFF
--- a/packages/custom-webpack/src/custom-webpack-builder.spec.ts
+++ b/packages/custom-webpack/src/custom-webpack-builder.spec.ts
@@ -50,45 +50,48 @@ const customWebpackFunctionObj = {
 	}
 };
 
-function createConfigFile(fileName: string) {
-	jest.mock(`${__dirname}/${fileName}`, () => customWebpackConfig, {virtual: true});
-}
-
-function createConfigFunction(fileName: string) {
-	jest.mock(`${__dirname}/${fileName}`, () => customWebpackFunction, {virtual: true})
+function createConfigFile(fileName: string, content: any) {
+	jest.unmock(`${__dirname}/${fileName}`);
+	jest.mock(`${__dirname}/${fileName}`, () => content, {virtual: true});
 }
 
 describe('CustomWebpackBuilder test', () => {
 	it('Should load webpack.config.js if no path specified', () => {
-		createConfigFile(defaultWebpackConfigPath);
+		createConfigFile(defaultWebpackConfigPath, customWebpackConfig);
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {}, baseWebpackConfig, {});
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, undefined, undefined);
 	});
 
 	it('Should load the file specified in configuration', () => {
 		const fileName = 'extra-webpack.config.js';
-		createConfigFile(fileName);
+		createConfigFile(fileName, customWebpackConfig);
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {path: fileName}, baseWebpackConfig, {});
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, undefined, undefined);
 	});
 
 	it('Should pass on merge strategies', () => {
-		createConfigFile(defaultWebpackConfigPath);
+		createConfigFile(defaultWebpackConfigPath, customWebpackConfig);
 		const mergeStrategies: MergeStrategies = {'blah': 'prepend'};
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {mergeStrategies}, baseWebpackConfig, {});
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, mergeStrategies, undefined);
 	});
 
 	it('Should pass on replaceDuplicatePlugins flag', () => {
-		createConfigFile(defaultWebpackConfigPath);
+		createConfigFile(defaultWebpackConfigPath, customWebpackConfig);
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {replaceDuplicatePlugins: true}, baseWebpackConfig, {});
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, undefined, true);
 	});
 
 	it('Should execute custom function on configuration', () => {
-		const fileName = 'custom-webpack.config.js';
-		createConfigFunction(fileName);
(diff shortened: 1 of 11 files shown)


mistic100:
Missed some code and tests :)

packages/custom-webpack/src/custom-webpack-builder.spec.ts
```diff
@@ -20,39 +20,75 @@ const customWebpackConfig = {
...
-});
\ No newline at end of file
+
+	it('Should execute custom function on configuration', () => {
+		const fileName = 'custom-webpack.config.js';
+		createConfigFunction(fileName);
+		const mergedConfig = CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {path: fileName}, baseWebpackConfig, {env: 'prod'});
+		 expect(mergedConfig).toEqual(customWebpackFunctionObj);
+	});
+});
```
meltedspark: Would be great to add another test checking that the second parameter was actually received in the function.

mistic100: testing that `prod: options.env === 'prod'` actually returns `prod: true` is not enough ?

mistic100: anyway, how would you test it explicitely ?

mistic100:
@meltedspark About E2E test, I took a look and honnestly I am lost. Your custom toolchain (sh scripts, multiple pack and copies, ...) is a bit complex. Could you do it ? It will be a lot quicker.

meltedspark:
@mistic100 Yeah, switching all this to plain JS is somewhere in my TODOs. I'll take this one, NP.

packages/custom-webpack/src/custom-webpack-builder.spec.ts
meltedspark: I meant that the preferred way is to have separate tests for separate functionality. Here you have a test that tests both that the function was applied and that the second parameter is received.  
Essentially if one of them breaks the test will fail and you'll have to understand which one has failed.  
The easiest way to test the "parameter was passed" thing is to pass a spy instead of a function (`jest.fn()`) and make sure that it `hasBeenCalledWith` the right parameters.

mistic100: done

packages/custom-webpack/src/custom-webpack-builder.spec.ts
```diff
@@ -50,45 +50,48 @@ const customWebpackFunctionObj = {
...
 	});
 
 	it('Should pass on replaceDuplicatePlugins flag', () => {
-		createConfigFile(defaultWebpackConfigPath);
+		createConfigFile(defaultWebpackConfigPath, customWebpackConfig);
 		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {replaceDuplicatePlugins: true}, baseWebpackConfig, {});
 		expect(WebpackConfigMerger.merge).toHaveBeenCalledWith(baseWebpackConfig, customWebpackConfig, undefined, true);
 	});
 
 	it('Should execute custom function on configuration', () => {
```
meltedspark: Should pass the build options to the webpack function

meltedspark: Should pass build options to the webpack config function

packages/custom-webpack/src/custom-webpack-builder.spec.ts
```diff
@@ -50,45 +50,48 @@ const customWebpackFunctionObj = {
...
-		const fileName = 'custom-webpack.config.js';
-		createConfigFunction(fileName);
-		const mergedConfig = CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {path: fileName}, baseWebpackConfig, {env: 'prod'});
-		 expect(mergedConfig).toEqual(customWebpackFunctionObj);
+		const spy = jest.fn((config, options) => config);
+		createConfigFile(defaultWebpackConfigPath, spy);
+		CustomWebpackBuilder.buildWebpackConfig(__dirname as Path, {}, baseWebpackConfig, {env: 'prod'});
```
meltedspark: Put {env: 'prod'} to variable

packages/custom-webpack/src/custom-webpack-builder.spec.ts
```diff
@@ -50,45 +50,48 @@ const customWebpackFunctionObj = {
...
+		expect(spy).toHaveBeenCalledWith(baseWebpackConfig, {env: 'prod'});
+	});
+
+	it('Should apply custom function on configuration', () => {
```
meltedspark: This test will fail if the second parameter won't be passed (while it checks whether the function was applied or not). So the applied function should ignore the second parameter (this way the test is truly isolated)

mistic100: Sorry I don't understand what you mean.

meltedspark: Try to break the logic you've implemented. In `packages/custom-webpack/src/custom-webpack-builder.ts` at line 17 don't pass the build options.  
You'll see that it will fail both tests, although one of them checks that the function has been applied. So the function is still being but the test that verifies that fails for completely different reason (the second parameter is not passed).  
When writing unit test you (ideally) want to be able to identify where is the problem by just looking at the name of the failing test.  
Is that clearer now?

packages/custom-webpack/src/custom-webpack-builder.spec.ts
meltedspark: This would be the right test description string

packages/custom-webpack/src/custom-webpack-builder.spec.ts
mistic100: I don't see the point on having such level of isolation for something that should work anyway (I would spend my entire life writing unit test if I had to apply this logic at work with our huge codebase).

It's done anyway.

meltedspark: Thanks

mistic100:
Thank you for your time.

meltedspark:
Thank you for your professionalism and dedication. The feat is available in version `7.4.1`.

just-jeb:
@all-contributors please add @mistic100  for code

PR #314, opened 2019-05-05
Support Ionic build

yazantahhan:
This adds the support for ionic build to use the custom angular builder.

Ionic uses their own class and schema. So I referenced them from their project, as anyone wants to use this builder, then it should have the @ionic dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+This is a fork from [meltedspark/angular-builders](https://github.com/meltedspark/angular-builders) that supports Ionic.
+You will be able to customize the Webpack build configuration with Ionic
+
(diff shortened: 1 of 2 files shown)


meltedspark:
I'm not sure I understand the purpose of this pull request. Why one can't just use their builder? What is the need to reference their schema and implementation from `@angular-builders`.

yazantahhan:
Thanks for the Reply.
So in order to use the extra webpack and customize the build configuration, I referenced their schema and implementation.

For now, I will keep the Repo as a fork to test it more.
Will close the PR for now.

Thanks

Issue #346, opened 2019-06-18
setupFilesAfterEnv property cannot be set. 


michaeljota:
**Describe the bug**

If I set the `setupFilesAfterEnv` property in `jest.config.js`, the test will start throwing with random errors.

**To Reproduce**
1. Setup Jest as is in readme.
2. Add `setupFilesAfterEnv` property pointing to a file. 

Repo:
[jest-error.zip](https://github.com/meltedspark/angular-builders/files/3303407/jest-error.zip)

**Expected behavior**

No errors should be throw. 

**Builder:**
 - Jest builder
 - version: 8.0.2

**Libraries**
 - `angular-devkit/build-angular`: ~0.800.0

michaeljota:
Seems to be about how lodash does the merging with merge function, as it doesn't concat arrays. lodash/lodash#2872. Maybe we could update this to concat the arrays values in the configurations by using mergeWith as the issue suggest?

meltedspark:
Interesting... I was sure I fixed this issue a long time ago. Would you mind checking if the issue exists on latest 7 version? It also supports Jest 24, probably I screwed something during transition to version 8...

michaeljota:
[jest-error-v7.zip](https://github.com/meltedspark/angular-builders/files/3306971/jest-error-v7.zip)

The error does reproduce in 7.4.4, with Angular CLI 7. (I setup a different environment as v7 only works with Angular 7).

meltedspark:
Well, you are right then, we should use `mergeWith`. Would you like to issue a PR? If yes, please bear in mind it should include at least unit test, but ideally unit test and e2e test.

michaeljota:
I modify the file locally. I'll try to make the PR, but I make no promises. However, with this changes the builder was able to load a file through the `setupFilesAfterEnv` property. 

```ts
function concatArrayProperties(objValue: unknown, srcValue: unknown) {
  if (Array.isArray(objValue)) {
    if (Array.isArray(srcValue)) {
      return objValue.concat(srcValue);
    }
    // This will make Jest to throw, as it is expecting an array, instead of silently fail.
    return srcValue;
  }
}

export const buildConfiguration = (defaultConfigResolver: DefaultConfigResolver,
  customConfigResolver: CustomConfigResolver) => 
  (projectRoot: Path,workspaceRoot: Path, configPath: string = 'jest.config.js') => {
    const globalDefaultConfig = defaultConfigResolver.resolveGlobal();
[...]

PR #347, opened 2019-06-23
fix: setupFilesAfterEnv property cannot be set (#346)

michaeljota:
This will allow to declare properties that exist in the default configuration as array, and are not mean to be override. This is done by declaring a whitelist of properties that should concat their values. By using a helper function with `mergeWith` we can concat only those properties.


meltedspark:
Looks good. Why do you think it makes sense to concat AST transformers as well? What if someone actually wants to override it?

michaeljota:
I think is better to prevent an undesired behavior. I understand that there may be a use case where you would actually like to replace this entirely, but as I see it, most of the time if you would like to add something on top of the default value, just as setupFilesAfterEnv.

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -1,8 +1,38 @@
 import { Path, resolve } from "@angular-devkit/core";
-import { merge } from 'lodash';
+import { mergeWith } from 'lodash';
 import { CustomConfigResolver } from "./custom-config.resolver";
 import { DefaultConfigResolver } from "./default-config.resolver";
 
+/**
+ * A whitelist of property names that are mean to be concat.
```
meltedspark: *meant

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -1,8 +1,38 @@
...
+  // From ts-jest config
+  'astTransformers'
+];
+
+/**
+ * This function checks witch properties should be concat. Early return will
+ * merge the data as lodash#merge would do it. 
+ */
+function concatArrayProperties(objValue: unknown, srcValue: unknown, property: string) {
+  if (!ARRAY_PROPERTIES_TO_CONCAT.includes(property)) {
```
meltedspark: Can be united

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -1,8 +1,38 @@
...
+    return;
+  }
+
+  if (!Array.isArray(objValue)) {
```
meltedspark: Can be united with the previous if statement

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -1,8 +1,38 @@
...
+    return;
+  }
+  
+  if (!Array.isArray(srcValue)) {
+    // This will make Jest to throw, as it is expecting an array, instead of silently fail.
+    return srcValue;
```
meltedspark: I'd throw an explicit error here: "Tried to concat non-array property". Moreover the behavior should be the same for object value as well. If I tried to merge array value into non-array value while I listed it in the ARRAY properties I probably mistaken.

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -24,12 +54,7 @@ export class JestConfigurationBuilder {
   buildConfiguration(projectRoot: Path, workspaceRoot: Path, configPath: string = 'jest.config.js'): any {
```
meltedspark: No need in default argument anymore, since you're using buildConfiguration function

packages/jest/src/jest-configuration-builder.ts
michaeljota: Sure. I miss that.

packages/jest/src/jest-configuration-builder.ts
michaeljota: But this would have an inconsistent behavior for the properties we don't concat, and replace instead. If we leave Jest to throw, it would throw the same error, regardless is a property we need to concat or not.

meltedspark: We don't want the error to be thrown if it's not a property we want to concat. We do want to throw an error if it is a property we want to concat but not an array. So the inconsistency is fine, what am I missing?

michaeljota: Nothing. Just that I though it would be better to let Jest to throw, as it is currently happening. But I'll update it.

packages/jest/src/jest-configuration-builder.ts
michaeljota: I update this as well. Thanks.

michaeljota:
A though about this. What if the user configure some of this on the angular.json and some of this on the jest.config? .-.

meltedspark:
It is not necessarily related to this pull request, but if you look [here](https://github.com/meltedspark/angular-builders/blob/master/packages/jest/src/index.ts) you'll see that the configuration is provided to Jest as one of the CLI arguments. So if an argument that already exists in configuration is provided as a separate CLI option the behavior would be the same as default Jest behavior. Need to verify that but I think it would override the config param.

packages/jest/src/jest-configuration-builder.ts
```diff
@@ -18,17 +18,12 @@ const ARRAY_PROPERTIES_TO_CONCAT = [
  * merge the data as lodash#merge would do it. 
  */
 function concatArrayProperties(objValue: unknown, srcValue: unknown, property: string) {
-  if (!ARRAY_PROPERTIES_TO_CONCAT.includes(property)) {
+  if (!ARRAY_PROPERTIES_TO_CONCAT.includes(property) || !Array.isArray(objValue)) {
```
meltedspark: I think of `objValue` is not an array it should throw as well, no? It still means that we're trying to concat a property which is not an array, meaning that we've put it in the list by mistake.

michaeljota: I think you are right. But that should be catch with unit testing. So, I will remove that from the code.

michaeljota: Updated.

michaeljota:
Sorry for the delay. I updated the PR to address the latest comment.

meltedspark:
@michaeljota It is still inconsistent. It's one of the two:
1. We say that we are sure that the properties we put in `ARRAY_PROPERTIES_TO_CONCAT` are in fact array properties (and we cover it with UT as you said). In this case there is no need to validate either source value or object value for being an array, because, well, we know that if it's in `ARRAY_PROPERTIES_TO_CONCAT` it is an array.
2. We are not sure about this. In this case there are two possible mistakes: 
    - We're trying to contact array to non-array property
    - We're trying to contact object to an array property
   Both mistakes are possible so we should validate both objValue and srcValue for being an array and throw an error if they are not.

In conclusion: we should either validate both values or none. Validating one of them gives us inconsistent behavior.

michaeljota:
@meltedspark We are validating what's coming from the user input, as we would be validating our own input twice, and this seems like a mistake.

michaeljota:
If you don't want to verify input from the user, then I'll updated. This is actually not necessary, but only made as the user should use the same input schema as Jest is intended.

michaeljota:
@meltedspark with the latest changes that you did, is this ready to merge? Thank you.

meltedspark:
Sure, just wanted to make sure it's fine by you. It is your PR after all.

michaeljota:
Oh! Thanks. Yes. I ask you to do something like what you did.

just-jeb:
@all-contributors please add @michaeljota for code

PR #504, opened 2019-09-12
Update README.md

sankety:
Added moduleNameMapper config option for applications using angular libraries as of latest version of angular. Without this config the test case will failed wherever the library is imported.


sankety:
> ulti-project support

Thanks @just-jeb 
Yes this is for multi project scenario.
You are right it should be under the "https://github.com/just-jeb/angular-builders/tree/master/packages/jest#multi-projects-workspace-support"
I dont see where exactly this snippet fits there, please add as you feel right.
But do these changes as it took me a while to figure out what was going on, so it will save some time for others.

PR #614, opened 2019-11-13
feat(custom-webpack): support exporting `Promise`

### LINKED ISSUE #274: Allow Config Function To Return Promise
Yes, I can't do something asynchronous in my config function.

Allow the config function to return a promise, e.g.

```typescript
module.exports = async (config, options) => {
  return config;
}
```

**Additional context**
@angular-builders/custom-webpack `v7.4.3`

arturovt:
Closes #274 

I had to refactor tests a little bit, because they weren't working correctly. See screenshot below :point_down: 

invalid-test

`devkit/build-angular` is able to process Webpack config that exports asynchronous factory. [See ExecutionTransformer signature](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/transforms.ts#L8).

Also now tests are run successfully :point_down: 

Снимок экрана от 2019-11-13 22-16-27

I mean that Jest throws exceptions on invalid expectations :)


just-jeb:
Actually when I assessed this feature I was thinking about supporting both async functions and Promise config. Meaning both ways (function and object) can be provided asynchronously. So that eventually we have 4 possible configuration options:
1. Config object
2. Promise that resolves to config object
3. Function that modifies the configuration
4. Async function that modifies the configuration.  
The second option is good for use cases when you have to instantly perform an async action (like read stuff from a file) and return partial config based on this operation. 
The change is quite small (you just need to await on the object) but it will totally close the gap.  

Also, documentation should be updated to reflect the changes. 

What do you think?

arturovt:
Probably I got your idea. Do you mean this:
```js
// config.js
module.exports = new Promise(r => { ... })
```
?

just-jeb:
Yes, exactly! Although in real life it will look like something like this I guess:
```js
module.exports = fs.readFile('my-version').then(version => ({
  //some-version-plugin-crap
}) 
```

arturovt:
Ok. I will update the PR accordingly, thus it will become `feat`.

packages/custom-webpack/src/custom-webpack-builder.ts
```diff
@@ -1,25 +1,61 @@
...
-        //TODO: support function returning promise
-        if (typeof customWebpackConfig === "function") {
-            return customWebpackConfig(baseWebpackConfig, buildOptions);
-        } else {
-            return mergeConfigs(baseWebpackConfig, customWebpackConfig, config.mergeStrategies, config.replaceDuplicatePlugins);
+        const configOrFactoryOrPromise: CustomWebpackConfig = require(`${getSystemPath(
+            root
+        )}/${webpackConfigPath}`);
+
+        if (configOrFactoryOrPromise instanceof Promise) {
```
just-jeb: I guess it's a matter of personal preference but I prefer the `async/await` syntax. Since you can await on non-promises (which just creates a promise of awaited object) the code becomes cleaner and shorter:
```typescript
async buildWebpackConfig(
        root: Path,
        config: CustomWebpackBuilderConfig,
        baseWebpackConfig: Configuration,
        buildOptions: any
    ): Promise<Configuration> {
        if (!config) {
            return baseWebpackConfig;
        }

        const webpackConfigPath = config.path || defaultWebpackConfigPath;
        const configOrFactoryOrPromise: CustomWebpackConfig = require(`${getSystemPath(
            root
        )}/${webpackConfigPath}`);

        if (typeof configOrFactoryOrPromise === 'function') {
            // That exported function can be synchronous either
            // asynchronous. Given the following example:
            // `module.exports = async (config) => { ... }`
            return configOrFactoryOrPromise(baseWebpackConfig, buildOptions);
        }

        const resolvedConfig = await configOrFactoryOrPromise;
        // In this case the user has exported a plain object, like:
        // `module.exports = { ... }`
        return mergeConfigs(
            baseWebpackConfig,
            resolvedConfig,
            config.mergeStrategies,
            config.replaceDuplicatePlugins
        );
    }
```

packages/custom-webpack/README.md
```diff
@@ -216,6 +216,27 @@ new CircularDependencyPlugin({
...
+const webpack = require('webpack');
+
+const readFile = util.promisify(fs.readFile);
+
+module.exports = readFile('./LICENSE', {
+  encoding: 'utf-8'
+}).then(license => ({
+  plugins: [new webpack.BannerPlugin(license)]
+}));
+```
```
just-jeb: I think it is worth to add a word or two about support of async config function as well (maybe in Cstuom webpack config function section)

packages/custom-webpack/src/custom-webpack-builder.ts
arturovt: I thought about this, but will need to fix all tests since the Webpack config will be always “gotten” asynchronously.

arturovt:
@just-jeb I messaged on you in Twitter, could you check? :D

PR #627, opened 2019-11-20
fix(custom-webpack): support exporting function in config written in TS

a1russell:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

I noticed that TypeScript support was recently added, so I tried it out. It didn't seem to work with my config, and I am pretty sure it's because I am exporting a function rather than an object. The code that was added to support TypeScript only seems to test for objects before my PR.

Also, apologies, I couldn't find any tests to modify to account for my fix. You'll have to point them out to me if they exist.

## What is the new behavior?

Exported function works with TypeScript config files.

```
[ ] Yes
[x] No
```

## Other information

I should also point out that I couldn't get `module.exports` syntax to work at all. I think that would require change to the tsconfig, e.g. [`esModuleInterop` or something as described on the Webpack documentation](https://webpack.js.org/configuration/configuration-languages/). I didn't try to fix that, though.

### DIFF
--- a/packages/custom-webpack/README.md
+++ b/packages/custom-webpack/README.md
@@ -316,7 +316,7 @@ module.exports = (config, options) => {
 Alternatively, using TypeScript:
 
 ```ts
-import { CustomWebpackBrowserSchema } from '@angular-builders/custom-webpack/dist';
+import { CustomWebpackBrowserSchema } from '@angular-builders/custom-webpack';
 import * as webpack from 'webpack';
 import * as pkg from './package.json';
 
--- a/packages/custom-webpack/README.md
+++ b/packages/custom-webpack/README.md
@@ -313,6 +313,24 @@ module.exports = (config, options) => {
 };
 ```
 
+Alternatively, using TypeScript:
+
+```ts
+import { CustomWebpackBrowserSchema } from '@angular-builders/custom-webpack/dist';
+import * as webpack from 'webpack';
+import * as pkg from './package.json';
+
+export default (config: webpack.Configuration, options: CustomWebpackBrowserSchema) => {
+  config.plugins.push(
+    new webpack.DefinePlugin({
+      APP_VERSION: JSON.stringify(pkg.version),
+    })
+  );
+
+  return config;
+};
+```
+
 It's also possible to export an asynchronous factory (factory that returns a `Promise` object). Let's look at the following example:
 
 ```js
--- a/packages/custom-webpack/src/custom-webpack-builder.ts
+++ b/packages/custom-webpack/src/custom-webpack-builder.ts
@@ -67,9 +67,5 @@ function resolveCustomWebpackConfig(path: string): CustomWebpackConfig {
   // `module.exports = { ... }`. And the second one is:
   // `export default { ... }`. The ESM format is compiled into:
   // `{ default: { ... } }`
-  if (typeof customWebpackConfig.default === 'object') {
-    return customWebpackConfig.default;
-  }
-
-  return customWebpackConfig;
+  return customWebpackConfig.default || customWebpackConfig;
 }
--- a/packages/custom-webpack/examples/full-cycle-app/angular.json
+++ b/packages/custom-webpack/examples/full-cycle-app/angular.json
@@ -39,7 +39,7 @@
           "configurations": {
             "production": {
               "customWebpackConfig": {
-                "path": "./func-webpack.config.js"
+                "path": "./func-webpack.config.ts"
               },
               "fileReplacements": [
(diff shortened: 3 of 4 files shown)


just-jeb:
@all-contributors please add @a1russel for bug and code.

just-jeb:
@all-contributors please add @a1russell for bug and code

arturovt:
@a1russell hey, that would be great if you'd used the original PR template, thanks!

> can you point out to the tests please?

Go to `packages/custom-webpack/examples/full-cycle-app`. See the `func-webpack.config.js` file there. Change its extension to `.ts` and also the path in `angular.json`.

> I should also point out that I couldn't get module.exports syntax to work at all. 

Could you provide a working example?

arturovt:
@a1russell change the commit name to
> fix(custom-webpack): support exporting function in config written in TS

to follow the conventional commits notation.

packages/custom-webpack/src/custom-webpack-builder.ts
```diff
@@ -67,7 +67,10 @@ function resolveCustomWebpackConfig(path: string): CustomWebpackConfig {
   // `module.exports = { ... }`. And the second one is:
   // `export default { ... }`. The ESM format is compiled into:
   // `{ default: { ... } }`
-  if (typeof customWebpackConfig.default === 'object') {
+  if (
```
arturovt: What do you think about this?
```js
return customWebpackConfig.default || customWebpackConfig;
```
Less code I guess but as you wish.

a1russell:
> @a1russell hey, that would be great if you'd used the original PR template, thanks!

Edited original description to add back template.

> > can you point out to the tests please?
> 
> Go to `packages/custom-webpack/examples/full-cycle-app`. See the `func-webpack.config.js` file there. Change its extension to `.ts` and also the path in `angular.json`.

Thanks. I'll give this a go.

> > I should also point out that I couldn't get module.exports syntax to work at all.
> 
> Could you provide a working example?

I was referring to the [syntax mentioned in the comment above the code I edited](https://github.com/just-jeb/angular-builders/blob/%40angular-builders/custom-webpack%408.4.0/packages/custom-webpack/src/custom-webpack-builder.ts#L67). I honestly am not sure what a working example would look like for that.

arturovt:
Could you edit commit name here?

image

You should have the `Edit` button here, click it and change it to:
> 
fix(custom-webpack): support exporting function in config written in TS

packages/custom-webpack/src/custom-webpack-builder.ts
a1russell: Done

just-jeb:
Need to update the documentation in order to reflect the changes.

a1russell:
What documentation would need to be updated, please?

just-jeb:
I'd say this is the most appropriate place: https://github.com/just-jeb/angular-builders/tree/master/packages/custom-webpack#custom-webpack-config-function. 
Just give an example with TypeScript at the end.

packages/custom-webpack/README.md
```diff
@@ -313,6 +313,25 @@ module.exports = (config, options) => {
 };
 ```
 
+Alternatively, using TypeScript :
```
arturovt:
```suggestion
Alternatively, using TypeScript:
```

packages/custom-webpack/README.md
```diff
@@ -313,6 +313,25 @@ module.exports = (config, options) => {
...
+
+
+```typescript
```
arturovt:
```suggestion
```ts
```

packages/custom-webpack/README.md
```diff
@@ -313,6 +313,25 @@ module.exports = (config, options) => {
...
+import { CustomWebpackBrowserSchema } from '@angular-builders/custom-webpack/dist';
```
arturovt: @just-jeb should this be used publicly?

just-jeb: @arturovt  I don't see any problem with that.

arturovt: Maybe we could expose it publicly, thus it won’t be imported from the dist folder?

just-jeb: Sure, up until TS support there was no need in exporting anything but now it totally makes sense.

just-jeb: Actually it is already exported, so we just need to set `"types": "dist/index.d.ts"` in `package.json` (I believe it should work event without it, it should be the default entry)

PR #633, opened 2019-12-02
feat(custom-webpack): implement `verbose` option

arturovt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

**This is just a draft PR (not completed implementation). You can leave comments/suggestions/requests already.**

### DIFF
--- a/packages/custom-webpack/src/custom-webpack-builder-config.ts
+++ b/packages/custom-webpack/src/custom-webpack-builder-config.ts
@@ -6,4 +6,10 @@ export interface CustomWebpackBuilderConfig {
   path?: string;
   mergeStrategies?: MergeStrategies;
   replaceDuplicatePlugins?: boolean;
+  verbose?: CustomWebpackBuilderVerboseConfig;
+}
+
+export interface CustomWebpackBuilderVerboseConfig {
+  properties?: string | string[];
+  serializationDepth: number;
 }
--- a/packages/custom-webpack/src/verbose-logger.ts
+++ b/packages/custom-webpack/src/verbose-logger.ts
@@ -1,0 +1,36 @@
+import { isString, get } from 'lodash';
+import { Configuration } from 'webpack';
+import { LoggerApi } from '@angular-devkit/core/src/logger';
+
+import { CustomWebpackBuilderVerboseConfig } from './custom-webpack-builder-config';
+
+export interface VerboseLogger {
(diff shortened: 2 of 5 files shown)


packages/custom-webpack/src/custom-webpack-builder-config.ts
```diff
@@ -6,4 +6,10 @@ export interface CustomWebpackBuilderConfig {
   path?: string;
   mergeStrategies?: MergeStrategies;
   replaceDuplicatePlugins?: boolean;
+  verbose?: CustomWebpackBuilderVerboseConfig;
+}
+
+export interface CustomWebpackBuilderVerboseConfig {
+  options?: string | string[];
```
just-jeb: What are `options`? I'd prefer the interface to be very explicit with the only allowed properties.

packages/custom-webpack/src/verbose-logger.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(verbose: CustomWebpackBuilderVerboseConfig | undefined) {
+    this.setLoggableOptions(verbose);
+  }
+
+  logFinalConfig(config: Configuration): void {
+    for (const option of this.options) {
+      const value = get(config, option);
+
+      if (value) {
+        console.log(value);
```
just-jeb: We should be using here Angular logger.

arturovt: Okay.

packages/custom-webpack/src/custom-webpack-builder-config.ts
arturovt: Okay.

packages/custom-webpack/src/verbose-logger.ts
```diff
@@ -0,0 +1,32 @@
+import { isString, get } from 'lodash';
+import { Configuration } from 'webpack';
+
+import { CustomWebpackBuilderVerboseConfig } from './custom-webpack-builder-config';
+
+export class VerboseLogger {
```
just-jeb: I prefer switching to functions (in this case closure with options). 
Didn't have to to make the change project-wise but it is planned. So any new code is better to be written in functional paradigm rather than OOP.

PR #676, opened 2020-01-30
chore(jest): update to jest@25

just-jeb:


kroeder:
Anything blocking this PR? I'm wondering if it is save to use jest 25 in projects as well.

Do you need any help for jest 25 support?

just-jeb:
As far as I'm concerned there is nothing that blocks you from using Jest 25 with the latest version of this builder. `jest-preset-angular` didn't have any significant changes during the [upgrade](https://github.com/thymikee/jest-preset-angular/commit/8475b05c39f4fb47ed22f50c1c17da09998170b1) to Jest 25 so it should be fine.  
I guess this is also the reason for this PR not being worked on - it simply not in priority right now.  
However, the contributions are always welcome, so if you feel like completing this PR please go ahead.  
The intention of this PR is to upgrade all the Jest dependencies (`devDependencies` for the tests of the builders, `dependencies` in the example apps and `peerDependencies` in the Jest builder).

matheo:
Hi @just-jeb 
I just opened https://github.com/just-jeb/angular-builders/pull/765 because we get a lot of
```
ts-jest[versions] (WARN) Version 25.5.4 of jest installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=24.0.0 <25.0.0). Please do not report issues in ts-jest if you are using unsupported versions.
```
and that's because an old `ts-jest` dependency in `jest-preset-angular`.

Perhaps I created it for the `8.x.x` branch because we need Angular 8 support, I guess it could be cherry-picked to the other branches too.

Issue #687, opened 2020-02-21
Support TS for indexTransform


at-vinguyen:
Now, I can't import **_ts file_** in `index-transform.js`, see the code snippet below (just in case):
```js
import { environment } form 'environment.ts'
module.exports = (targetOptions, indexHtml) => {
  // use environment info to do some things
}
```
So I think it's better if TS is supported for indexTransform then we can use TypeScipt great features if needed.

just-jeb:
[Related](https://github.com/just-jeb/angular-builders/issues/681#issuecomment-588302447)

vixnguyen:
@just-jeb 
Can you allow me to create a new PR to solve this issue?

just-jeb:
Sure, go for it!

vixnguyen:
It seems I don't have permission to push my branch 🤔

just-jeb:
You should create a fork from this repo, implement the changes there and create a PR from the fork.

vixnguyen:
Already finished. Please help me to check 🎉

just-jeb:
Closed by #689

PR #689, opened 2020-02-24
[Enhancement] Support TS for indexTransform

vixnguyen:
## PR Checklist

## PR Type

- Build related changes

Issue Number: #687 

- No


packages/custom-webpack/src/common.ts
```diff
@@ -23,6 +23,14 @@ export const indexHtmlTransformFactory: (
   context: BuilderContext
 ) => IndexHtmlTransform = ({ indexTransform }, { workspaceRoot, target }) => {
   if (!indexTransform) return null;
+  if (indexTransform.endsWith('.ts')) {
```
just-jeb: Should be externalized to a separate function (in this file) and reused [here](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/src/custom-webpack-builder.ts#L55-L62)

packages/custom-webpack/README.md
```diff
@@ -362,7 +362,7 @@ module.exports = async config => {
 **_Requires `@angular-devkit/build-angular@0.801` and `@angular-builders/custom-webpack@8.1.0`._**
 
 Since Angular 8 `index.html` is not generated as part of the Webpack build. If you want to modify your `index.html` you should use `indexTransform` option.  
-`indexTransform` is a path (relative to workspace root) to a `.js` file that exports transformation function for `index.html`.  
+`indexTransform` is a path (relative to workspace root) to a `.js` or `.ts` file that exports transformation function for `index.html`.
```
just-jeb: Please add a TS example (similarly to TS example for webpack.config file).

packages/custom-webpack/src/common.ts
vixnguyen: I created a function in `custom-webpack-builder` and reused here to avoid an overlap importing. If not OK, I think it better to write a new file 🤔

packages/custom-webpack/src/custom-webpack-builder.ts
```diff
@@ -51,15 +51,25 @@ export class CustomWebpackBuilder {
...
 }
 
-function resolveCustomWebpackConfig(path: string): CustomWebpackConfig {
-  if (path.endsWith('.ts')) {
+/**
+ * check for TS node registration
+ * @param file: string file name or file directory are allowed
+ * @todo tsNodeRegistration: require ts-node if file extendsion is TypeScript
+ */
+export function tsNodeRegister(file: string = '') {
```
just-jeb: Yes, probably externalizing it into a separate file is a good idea

packages/custom-webpack/examples/full-cycle-app/index-html.transform.ts
```diff
@@ -0,0 +1,7 @@
+export default (targetOptions, indexHtml) => {
```
just-jeb: Since it's a TS example we should showcase the types usage here.  
Please import the relevant types from `@angular-builders/custom-webpack` (if needed then reexport from `@angular-devkit`).

vixnguyen: In my opinion, use static type for `indexHtml` is good, but the other one is unnecessary, try to import a type and use for readonly obj just make our code more complicated. So, in that case, I think write document(readme) for it is enough.

just-jeb: I disagree, having a type for these options is very helpful and gives you the exact set of options you can use. Otherwise you should go back and forth to the documentation (given it's up to date).  
If it's a problem I can add the type myself.

vixnguyen: Haha. It's no problem to me 😄 , just discuss and raise my concerns. Moreover, you are the owner, you can ask me to do everything because I happy to contribute to your package and willing to follow your rules. So don't talk to me that way `If it's a problem I can add the type myself.`.

vixnguyen: And one more concern, I think `indexHtml` is more important than `targetOptions`, so it should be the first param, some cases we no need to use `targetOptions`. Is it possible to have a breaking change in the next release?

vixnguyen:
@just-jeb Please help me to check again. Thanks!

just-jeb:
@all-contributors, let's add @vixnguyen for his contributions to the code base.

just-jeb:
@all-contributors please add @vixnguyen code.

vixnguyen:
@just-jeb 
`itwcw` stand for `Index Transform without Custom Webpack`

PR #718, opened 2020-04-08
ci: setup GitHub Actions

arturovt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

Issue Number: #639

```
[ ] Yes
[x] No
```


.github/workflows/angular-builders.yml
```diff
@@ -0,0 +1,46 @@
...
+          key: ${{ runner.os }}-node-modules
+          restore-keys: ${{ runner.os }}-node-modules
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.12
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install dependencies
+        if: steps.node-modules.outputs.cache-hit != 'true'
```
just-jeb, line 31: I think it's problematic. This way you don't check that the cache is up to date. 
In all the `node_modules` examples they [use](https://github.com/actions/cache/blob/master/examples.md#node---yarn) lock files as a hash.  
We don't have lock files, so will happen here is essentially "use the old cache even if newer versions exist".  
I think the solution for us is to remove the `if`. Restore `node_modules` in any case and then run `yarn` to update all the dependencies matching the scope in `package.json`.

.github/workflows/angular-builders.yml
```diff
@@ -0,0 +1,46 @@
...
+    strategy:
+      fail-fast: true
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/cache@v1
+        id: node-modules
+        with:
+          path: node_modules
```
just-jeb, line 21: We should probably use multi-path feature for `master` as described [here](https://github.com/actions/cache/blob/master/examples.md#node---lerna).

.github/workflows/angular-builders.yml
just-jeb, line 31: I think it's problematic. This way you don't check that the cache is up to date. 
In all the `node_modules` examples they [use](https://github.com/actions/cache/blob/master/examples.md#node---yarn) lock files as a hash.  
We don't have lock files, so will happen here is essentially "use the old cache even if newer versions exist".  
I think the solution for us is to remove the `if`. Restore `node_modules` in any case and then run `yarn` to update all the dependencies matching the scope in `package.json`.

PR #765, opened 2020-06-21
Add support for jest@25 in v8.x.x

matheo:
## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

Support `jest v25` for Angular 8 websites.

```
[ ] Yes
[x] No
```


just-jeb:
Hey @matheo, thanks for the PR. Correct me if I'm wrong, but it is a breaking change. It does change the underlying version of `jest-preset-angular` to the next major version and also makes some changes to the Jest config.   
If we introduce this change as a minor change to version 8 of the builder it will break a lot of builds and will force people to move to Jest 25, which is undesirable.   
The recommended way here is to upgrade to Angular 9 (and corresponding version of the builder), but I understand it is not always possible. Maybe you can create a fork version?
What do you think?

matheo:
@just-jeb you're right, I forgot that most of the projects uses `^` instead `~`.

I was considering the fork already, I will have to do that in the meantime.
Thanks!

just-jeb:
@matheo You're welcome. If you need any help feel free to ask here or on Slack. Also, it would be nice if you shared the link to the fork once it's published for the others.  
The real solution though is to move to version 9.

matheo:
@just-jeb I've tried to update `jest-preset-angular` to `8.0.0` and `8.2.0` with no success. I get
image
on all my tests and they fail :(

I've tried different possibilities with `jest` 24 and 25 checking no mixed dependencies and NOPE.
I won't fork then and I will keep the working setup :P

Issue #816, opened 2020-08-20
Cannot find module packag.json


muuvmuuv:
## Describe the Bug

I would like to use a custom webpack typescript config with the define plugin but cannot import package.json. Which tsconfig is taken by angular-builders for the webpack config compilation? I tried adding that file to every webpack config without success.

```
webpack.config.ts:7:22 - error TS2732: Cannot find module './package.json'. Consider using '--resolveJsonModule' to import module with '.json' extension

7 import * as pkg from './package.json'
```

## Minimal Reproduction

1. Create a start
2. Include `@angular-builders/custom-webpack:browser`
3. Create a `webpack.config.ts`
4. Add package.json as a module
5. Run `npm run build`

## Environment

<pre><code>
Libs
- @angular/core version: 10.0.10
- @angular-devkit/build-angular version: 0.1000.6
- @angular-builders/custom-webpack version: 10

For Tooling issues:
- Node version: 12.18.0
- Platform: latest macOS

Others:

</code></pre>

muuvmuuv:
I fixed it by adding `resolveJsonModule` to the `ts-node` config in `utils`. Additionally I created this webpack tsconfig to make sure it does error (if it errors) before the production build in our pipeline. Maybe this should be added as an option to pass a custom tsconfig to `ts-node` our by default take `tsconfig.webpack.json` if it exists.

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "baseUrl": "./",
    "outDir": "./out-tsc/webpack",
    "strictFunctionTypes": false,
    "resolveJsonModule": true,
    "declaration": false,
    "module": "ESNext",
    "moduleResolution": "node",
    "target": "ES2015",
    "lib": [
      "ESNext"
    ],
    "types": [
      "node"
    ]
  },
  "include": [
    "webpack.*.ts"
  ]
}
```

muuvmuuv:
Oh wait, adding this to `ts-node` still errors for `@angular-builders/custom-webpack:karma`.

I will try to setup a repl with all cases and the possible fix.

muuvmuuv:
Repl: https://github.com/muuvmuuv/angular-builders-issues-816

just-jeb:
Thanks for reporting, I'll take a look. Would you like to create a PR with a fix for `custom-webpack:browser`?
Have you tried using `require` instead of `import`?

muuvmuuv:
I may not find any time this week but will put it on my schedule next week :)

[...]

PR #824, opened 2020-09-02
feat(custom-webpack): add project to ts node registration

muuvmuuv:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #816

## What is the new behavior?

This adds `project` to `ts-node` configuration. This way it is possible to e.g. add `resolveJsonModule` and use JSON files in the webpack esm config.

```
[ ] Yes
[x] No
```

## Other information

@just-jeb I noticed that when running `yarn format` a lot is getting formatted. Maybe you should run that on the master branch once and include it in an husky hook?

closes: #816


packages/custom-webpack/src/utils.ts
```diff
@@ -1,15 +1,24 @@
...
-      compilerOptions: {
-        module: 'commonjs',
-      },
-    });
+export function tsNodeRegister(file = '', root = process.cwd()) {
+  if (!file.endsWith('.ts')) {
+    return;
   }
+
+  const tsConfigPath = resolve(root, 'tsconfig.app.json');
```
just-jeb, line 14: Actually `tsConfig` path should be taken from the config as well (we cannot assume it's always `tsconfig.app.json`).
https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/examples/full-cycle-app/angular.json#L28

packages/custom-webpack/src/utils.ts
```diff
@@ -1,15 +1,24 @@
...
+
+  const { register } = require('ts-node');
+
+  register({
+    project: tsConfigPath,
+    compilerOptions: {
```
just-jeb, line 20: If we provide the project then maybe we don't need the `compilerOptions` here?

just-jeb (commented):
Would you mind adding your scenario to e2e tests? You do this by adding a `package.json` import into [this](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/examples/full-cycle-app/extra-webpack.config.ts) file and adding `resolveJsonModule` flag to the tsconfig in the example project.

packages/custom-webpack/src/utils.ts
muuvmuuv: I am not quite sure if `project` overrides it but in one of my projects I have `"module": "ES2020",`.

just-jeb:
Also, the build fails [here](https://travis-ci.com/github/just-jeb/angular-builders/builds/182400468#L1025). I remember having this issue in the past (the file is transpiled twice because `ts-node` is registered twice). As far as I remember it happened because of two different versions of `ts-node` being installed, not sure why it happens here.

just-jeb:
> I noticed that when running yarn format a lot is getting formatted. Maybe you should run that on the master branch once and include it in an husky hook?

Yep, definitely [needs to be done](https://github.com/just-jeb/angular-builders/issues/826).

packages/custom-webpack/src/utils.ts
muuvmuuv: Can you have a look again. I have also fixed some typing issues I encountered while moving things.

muuvmuuv:
> Also, the build fails [here](https://travis-ci.com/github/just-jeb/angular-builders/builds/182400468#L1025). I remember having this issue in the past (the file is transpiled twice because `ts-node` is registered twice). As far as I remember it happened because of two different versions of `ts-node` being installed, not sure why it happens here.

I noticed running `yarn test` prints a huge amount of errors about zone missing async. Can you tell me how I could simulate the Travis build locally?

just-jeb:
It depends. If you want to run unit tests then it's enough to run `yarn test` in the package, but if you want to run E2Es then you should run `yarn ci` in the root. 
This will essentially build the packages and run the [ci script](https://github.com/just-jeb/angular-builders/blob/master/scripts/run-ci.sh), which, in turn, will setup local registry, publish the built packages, bootstrap example projects with the published versions and execute `ci` script in each one of the packages.  
If you want just to run the e2e for `custom-webpack` only you may as well pack it the built version with `npm pack`, install it manually in the example projects and run the `ci` script from within the `custom-webpack` folder.

muuvmuuv:
I really don't understand why webpack.ProvidePlugin fails. Now I tried to create a new project with `❯ ng new test --minimal --no-interactive --skip-git --skip-tests --defaults --skip-install && cd test && yarn`, add and link custom-webpack but then get:

```
Compiling @angular/core : es2015 as esm2015
Error: Error on worker #1: TypeError: ts.updateIdentifier is not a function
    at Object.generateSetClassMetadataCall (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/annotations/src/metadata.js:34:21)
    at InjectableDecoratorHandler.analyze (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/annotations/src/injectable.js:73:46)
    at NgccTraitCompiler.TraitCompiler.analyzeTrait (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/transform/src/compilation.js:346:40)
    at analyze (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/transform/src/compilation.js:298:58)
    at _loop_1 (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/transform/src/compilation.js:320:21)
    at NgccTraitCompiler.TraitCompiler.analyzeClass (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/src/ngtsc/transform/src/compilation.js:326:35)
    at NgccTraitCompiler.analyzeFile (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/analysis/ngcc_trait_compiler.js:48:26)
    at DecorationAnalyzer.analyzeProgram (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/analysis/decoration_analyzer.js:140:39)
    at Transformer.analyzeProgram (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/packages/transformer.js:130:57)
    at Transformer.transform (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/packages/transformer.js:76:27)
    at ClusterMaster.onWorkerMessage (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/execution/cluster/master.js:195:27)
    at /Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/execution/cluster/master.js:55:95
    at ClusterMaster.<anonymous> (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/execution/cluster/master.js:293:57)
    at step (/Users/marvinheilemann/Development/angular-builders/node_modules/tslib/tslib.js:140:27)
    at Object.next (/Users/marvinheilemann/Development/angular-builders/node_modules/tslib/tslib.js:121:57)
    at /Users/marvinheilemann/Development/angular-builders/node_modules/tslib/tslib.js:114:75
    at new Promise (<anonymous>)
    at Object.__awaiter (/Users/marvinheilemann/Development/angular-builders/node_modules/tslib/tslib.js:110:16)
    at EventEmitter.<anonymous> (/Users/marvinheilemann/Development/angular-builders/node_modules/@angular/compiler-cli/ngcc/src/execution/cluster/master.js:287:32)
    at EventEmitter.emit (events.js:315:20)
```

At least importing JSON works but without ProvidePlugin it is unusable for my project case. What do you say @just-jeb ? Should we merge this anyway or debug this further?

muuvmuuv:
Everything works fine except that the karma builder seems to have issues with ts-node. Idk how this can be possible since browser and karma uses the same function `buildWebpackConfig`. Do you have any ideas?

just-jeb:
You mean the `build` is passing but the `test` is failing on build step? [This](https://travis-ci.com/github/just-jeb/angular-builders/builds/183320733#L957) error is relevant for `test` only?  
They indeed use the same function to retrieve the config, but the processes are different. `ng build` runs `webpack` and `ng test` runs [Karma](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/karma/index.ts#L156). If I had to guess I'd say the problem is there.  
Maybe because Angular has a different tsconfig for tests? So it registers the tsnode twice (first in our builder with `app` config and then in `karma` process with `spec` config)?   
Try to register with `spec` config when you're in the test target.

muuvmuuv:
@just-jeb it just uses the `tsConfig` defined in the task like we talked earlier, so it takes app in browser and spec in tests. I wasn't aware that is runs Karma Server but makes sense ^^ I might not get some time this week but have some short timespan today I can take to look at it. It just seems that `ts-node` does not register properly.

muuvmuuv:
It seems like this is related: https://github.com/karma-runner/karma/pull/3274

If we can disable the `ts-node` implementation from Karma through a migration step we can use ours properly.

just-jeb:
We need to understand the root cause better. I mean, it worked before the change (just the part with `package.json` doesn't) but it stopped working after the change so we're probably ruining something here.  

There are some [ways](https://github.com/lorenwest/node-config/pull/458/files) to check (kinda) if someone has already registered `ts` extension or to [disable](https://github.com/TypeStrong/ts-node/issues/409#issuecomment-407228078) Karma register. As far as I remember I tried the replacement solution when I first encountered this issue but it didn't work (can't recall why, but probably because that Karma register runs first).     

We probably need add a protecting condition anyways, however, it won't help us here - if we don't register it with our config (including `.json` files) your Karma run will still fail.  

The question is why the double registration doesn't break the Karma build before your change and why it does break it after the change. I suspect it's because of different project but we need to check this.  

Once we understand what's happening here and add the protecting condition I think there is a solution to your problem.  
By default the `("ts-node").register` would take the config from the root `tsconfig.json` so I think if you add `resolveJsonModule` there it will work with the Karma build.

just-jeb:
@muuvmuuv hey, have you had any progress on that?

muuvmuuv:
Hey @just-jeb, no sorry. I haven't had time for this or will be the next week. I need to finish a release.

muuvmuuv:
I am working on it but its blocked as long as Karma does not remove the ts-node requirement. Everything seems to work with latest beta. I am updating the examples currently and reducing their size.

muuvmuuv:
@just-jeb yeah! Seems like the only tests that are now failing or those e2e karam tests which will hopefully fixed after karma removes that hardcoded requirement stuff.

just-jeb:
It's been open for almost two years, why do you think it will be merged soon?

just-jeb:
Btw, #879 seems pretty similar to this and the tests there pass. Maybe it's not the Karma after all? Maybe it's the wrong tsconfig path or something (from a quick glance that's the difference between the two projects).  
How about you guys join the forces and create a single state of the art PR 😄?

just-jeb:
@all-contributors please add @muuvmuuv for code.

muuvmuuv:
@just-jeb you are kind of right. It is a mix of karma and angular dev-kit I think. I downloaded his project and tested it locally against his and my version. Same results. After diffing every change, I saw that he is adding "node" as a typing in the app config. And hey, that is where the error is coming from. Somehow only in the case of protractor it is checking types here. This is an antipattern as well because app is a browser not node build so this should be avoided. I am currently investigating what is causing this but it seems like the webpack config is passed to the angular compiler.

muuvmuuv:
IMO: since protractor gets more and more deprecated (in favor of cypress which we should add someday) in the angular world we shouldn't take much care about it, so I would just include node as typings for the succeeding of the tests. 

Anyway its confusing that the Angular team does not allow any custom tsconfig for that build target since karma also got one which can include node.

just-jeb:
Actually webpack (which is what matters here, since we're talking about builders) is in `node` world, not browser. Conceptually mixing app ts config and build ts config is wrong, but it's much simpler to use the same config I guess.

just-jeb:
Will be merged as part of #879

Issue #696, opened 2020-03-15
Option to disable global-mocks on jest builder


tonivj5:
## Describe the Bug

I'm having problems with the [global mocks](https://github.com/just-jeb/angular-builders/blob/master/packages/jest/src/jest-config/global-mocks.ts) configured, I would like to disable them. Could be posible to provide an option in `angular.json` to disable it?

## Minimal Reproduction

I think it isn't needed

## Expected Behavior

And option to disable the [mocks applied](https://github.com/just-jeb/angular-builders/blob/master/packages/jest/src/jest-config/global-mocks.ts)

## Environment

<pre><code>
Libs
- @angular/core version: latest
- @angular-devkit/build-angular version: latest
- @angular-builders/jest version: latest

For Tooling issues:
I think it isn't needed

Others:
None

just-jeb:
You don't need a special option for this. The global mocks can be easily overwritten if you add your own setup file to `setupFilesAfterEnv` and in this file redefine the window props.

tonivj5:
Ihave my own setup file with `setupFilesAfterEnv`, but they are not being overwritten, I think they are [concatenated](https://github.com/just-jeb/angular-builders/blob/master/packages/jest/src/jest-configuration-builder.ts#L11)

just-jeb:
The setup files array is concatenated, you're right, but that only means that the setup files run in order one after another. Which means if you redefine window properties in your setup file they should get overwritten even if the default setup file defined them before.

tonivj5:
> The setup files array is concatenated, you're right, but that only means that the setup files run in order one after another. Which means if you redefine window properties in your setup file they should get overwritten even if the default setup file defined them before.

That's exactly what I want to do, but I want to avoid the redefinetion of window's properties. I don't want to patch `window` and leave it as it's (prev to `global-mocks.ts` applied). I want this behavior because actually some patches are not necessary anymore...

just-jeb:
Well if some of the patches are not necessary anymore we probably should change it in the builder. Let's discuss.

tonivj5:
Ok, here you have a list of patches that I think don't be necessary anymore:
-  [this patch](https://github.com/just-jeb/angular-builders/blob/master/packages/jest/src/jest-config/global-mocks.ts#L3-L14) ==> [*storage are full implemented](https://github.com/jsdom/jsdom/blob/master/Changelog.md#11120) 
[...]

PR #843, opened 2020-09-29
fix(jest): remove local storage mock

zauni:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently the Jest builder is adding a mock for local storage and session storage. But this mock has missing features like the `length` property or a `keys` method.
This breaks unit tests which are using the localForage library because it uses the `length` property and `keys` method.

Issue Number: N/A

## What is the new behavior?

We remove the mock altogether because as the builder is requiring Jest >= 26 which requires the newest jsdom version (16) which itself has its own implementation of local storage and session storage. The implementation of jsdom has all missing features described above, so we don't need the mock anymore.

Jest 26 has jsdom 16: https://github.com/facebook/jest/releases/tag/v26.0.0
jsdom has a local storage and session storage implementation since 11.12.0: https://github.com/jsdom/jsdom/blob/master/Changelog.md#11120

```
[ ] Yes
[X] No
```

I would say it is not a breaking change because the implementation of the local storage and session storage of jsdom is the same as the current mock but with a more complete API.

### DIFF
--- a/packages/jest/src/jest-config/global-mocks.ts
+++ b/packages/jest/src/jest-config/global-mocks.ts
@@ -1,17 +1,5 @@
 global['CSS'] = null;
 
-const mock = () => {
-  let storage = {};
-  return {
-    getItem: (key: string) => (key in storage ? storage[key] : null),
-    setItem: (key: string, value: any) => (storage[key] = value || ''),
-    removeItem: (key: string) => delete storage[key],
-    clear: () => (storage = {}),
-  };
-};
-
-Object.defineProperty(window, 'localStorage', { value: mock() });
-Object.defineProperty(window, 'sessionStorage', { value: mock() });
 Object.defineProperty(document, 'doctype', {
   value: '<!DOCTYPE html>',
 });


just-jeb:
Hey @zauni, thanks for this PR. 
We've discussed this already in #696, and it indeed seems like it can be removed, along with a couple of other mocks.   
Let's just make sure it's not a breaking change. [Here](https://github.com/just-jeb/angular-builders/issues/696#issuecomment-602700383) is my previous analysis, we need to make sure that with Jest 26 our users build won't break upon mocks removal.

just-jeb:
@all-contributors please add @zauni for code.

zauni:
Hi @just-jeb thanks for commenting :) As the jest builder is already depending on Jest >= 26 the removal of the localStorage and sessionStorage should not be a breaking change, because the API of the mock and the jsdom implementation would not be different.

The removal of the "getComputedStyle" is possible, but would be a breaking change, because if someone is relying on the specific return value, then we would break his/her test. So maybe in a major version.

The removal of the document.body.style.transform is risky, because JSDOM 16 returns something different than the mock. The same goes for the doctype and the global CSS object. So maybe we should leave them in.

And matchMedia should be left in the mocks because JSDOM has no matchMedia implementation at the moment.

just-jeb:
@zauni OK, thanks for the thorough response and patience. I'm merging it now, you're welcome to test it in the next beta version.

just-jeb (approved):

Issue #841, opened 2020-09-24
Add Experimental Webpack 5 Support


xiaoxiangmoe:
There is an experimental support for Webpack v5 in the Angular CLI by https://github.com/angular/angular-cli/pull/18820, and it was released in 11.0.0-next.3

Can you add support for webpack v5 in @angular-builders/custom-webpack ?

Thank you very much!

just-jeb:
Hey @xiaoxiangmoe,
The latest version of this builder is to be used with Angular 10, most probably it won't work with Angular 11.  
We need to wait till Angular 11 is out and then release it with a new version of `@angular-builders`.  
That being said, even then the support is not promised. `@angular-builders` use [webpack-merge](https://github.com/survivejs/webpack-merge) for merging the configs and it seems that it doesn't have official support for Webpack 5 yet.

ydmitry:
@just-jeb 
Hello, I did a PoC maybe a month ago, with Angular CLI fork (it is released now) and with custom webpack builder. Everything worked for me, just was need to do 1-3 changes. Can create a PR

BenRacicot:
@ydmitry Awesome stuff! I'd love to see what you did and how you did it. 
I just setup a `11.0.0-next.5` project and am hoping to dig into module federation. Any guidance you may have would be so very appreciated.

just-jeb:
@ydmitry Hey, PRs are always welcome. Not sure it will be very helpful now, since current builder version explicitly depends on Angular Devkit 10, but it for sure will be a highly demanded feature in v11.

ydmitry:
OK, will try to do this weekends, actually it works with Angular in host mode. I assume need to fix "remote" mode

ydmitry:
Created a demo, everything is works when add "resolutions". Upgrading `peerDependencies` in custom webpack builder might be helpful, no need to other code changes on lib side.

Here is the demo:
https://github.com/ydmitry/angular11-webpack-module-federation-custom-webpack-builder

Had to add following resolutions:
```
"resolutions": {
    "webpack": "5.4.0",
    "@angular-devkit/build-angular": "~0.1100.0-rc.2",
    "@angular-devkit/architect": "~0.1100.0-rc.2",
    "@angular-devkit/core": "~11.0.0-rc.2",
    "webpack-cli": "4.0.0"
  }
```

qortex:
> Had to add following resolutions:
> 
> ```
> "resolutions": {
>     "webpack": "5.4.0",
[...]

PR #853, opened 2020-11-08
feat(webpack5): simplify dependencies resolution when opt-in

ydmitry:
Should help with https://github.com/just-jeb/angular-builders/issues/841 , so to opt-in to webpack 5 need to write only resolution for webpack 5

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/custom-webpack/package.json
+++ b/packages/custom-webpack/package.json
@@ -37,9 +37,9 @@
   },
   "builders": "builders.json",
   "dependencies": {
-    "@angular-devkit/architect": ">=0.1000.0 < 0.1100.0",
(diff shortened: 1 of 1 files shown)


just-jeb:
Hmm we actually can't do that, because builders work with specific version of Angular and major version can break builder's behavior.  
You can see how the changes that were introduced in #874. 
But I guess #874 solves the problem right? Can this PR be closed then?

Issue #749, opened 2020-05-28
@angular-builders/custom-webpack:karma fails compiling TypeScript with activated strictNullChecks


Markus-Ende:
## Describe the Bug

When a custom webpack config is provided written in TypeScript, @angular-builders/custom-webpack:karma throws an error if strictNullChecks are activated.

## Minimal Reproduction

https://github.com/Markus-Ende/custom-webpack-builder-strictnullchecks-issue

1. Clone & install repo
2.  run `npm test`.

(Deactive strictNullChecks in tsconfig.json to see, that the Error does only occur when activated)

## Expected Behavior

Karma tests should run normally.

## Error output

```
An unhandled exception occurred: ⨯ Unable to compile TypeScript:
webpack-config/extra-webpack.config.ts:2:54 - error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'Function'.

2 export default (config) => __awaiter(void 0, void 0, void 0, function* () {
                                                       ~~~~~~

See "/tmp/ng-Hi3EUl/angular-errors.log" for further details.
```

## Environment

<pre><code>
Libs
- @angular/core version: 9.1.9
- @angular-devkit/build-angular version: 0.901.7
- @angular-builders/custom-webpack version: 9.1.0

For Tooling issues:
- Node version: v12.16.1
- Platform:  Linux

</code></pre>

just-jeb:
Is this just Karma issue? No problems with `ng build` and `ng serve`?

Markus-Ende:
Yes, just Karma.

just-jeb:
Ok. A few findings:
1. The problem is in `async` keyword - once it is removed the problem goes away. So as a workaround you can just use plain `Promise` syntax. 
2. I can disable the type checking for the config file by passing `transpileOnly` option [into `register`](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/src/utils.ts#L9). Not sure it's the right decision though (it is basically your only indication for type errors in your config).
3. Another solution would be to have a separate `tsconfig.json` (built in the builder) for config transpilation. Not sure this is the right decision either, since I do want people to have control of it. 
4. I'd probably go with `#2` or `#3` if it was happening with all the commands. But it only happens with Karma and I don't have any idea why. So I think it's important to understand how Karma flow is different from the others before we take any actions.

Markus-Ende:
[...]

Issue #607, opened 2019-10-31
Add module mappings to projects in multi-project workspaces 


just-jeb:
Related to #504

just-jeb:
Closing this enhancement request — it has been open for 3+ years with no progress or community traction. If this is still relevant to you, please feel free to open a fresh issue with up-to-date context.

PR #879, opened 2020-11-18
Use application tsconfig file for indexTransform and customWebpackConfig

michelherv:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The application's `tsconfig` is ignore. Accordingly, `indexTransform` and `customWebpackConfig` cannot use project configuration for custom modules for example

Issue Number: #607 #749 

## What is the new behavior?
The application's `tsconfig` is now use. `indexTransform` and `customWebpackConfig` can now use all the application resources.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/custom-webpack/examples/full-cycle-app/angular.json
+++ b/packages/custom-webpack/examples/full-cycle-app/angular.json
@@ -50,7 +50,6 @@
               "optimization": true,
               "outputHashing": "all",
               "sourceMap": false,
-              "extractCss": true,
               "namedChunks": false,
               "aot": true,
               "extractLicenses": true,
--- a/packages/custom-webpack/examples/full-cycle-app/angular.json
+++ b/packages/custom-webpack/examples/full-cycle-app/angular.json
@@ -18,7 +18,10 @@
           "builder": "@angular-builders/custom-webpack:browser",
           "options": {
             "customWebpackConfig": {
-              "path": "./extra-webpack.config.ts"
+              "path": "./extra-webpack.config.ts",
+              "mergeRules": {
+                "plugins": "prepend"
+              }
             },
             "indexTransform": "./index-html.transform.js",
             "outputPath": "dist/full-cycle-app",
--- a/packages/custom-webpack/package.json
+++ b/packages/custom-webpack/package.json
@@ -42,6 +42,7 @@
     "@angular-devkit/core": "^11.0.0",
     "lodash": "^4.17.15",
     "ts-node": "^9.0.0",
+    "tsconfig-paths": "^3.9.0",
     "webpack-merge": "^5.7.3"
   }
 }
--- a/packages/custom-webpack/README.md
+++ b/packages/custom-webpack/README.md
@@ -261,35 +261,35 @@ The following properties are available:
    In more complicated cases you'd probably want to [use a function](#custom-webpack-config-function) instead of an object.
 - `mergeRules`: webpack config merge rules, as described [here](https://github.com/survivejs/webpack-merge#mergewithrules). Defaults to:
 
-```ts
-{
-  module: {
-    rules: {
-      test: "match",
-      use: {
-        loader: "match",
-        options: "merge",
+  ```ts
+  {
+    module: {
+      rules: {
+        test: "match",
+        use: {
+          loader: "match",
+          options: "merge",
+        },
       },
     },
-  },
-};
-```
+  };
+  ```
 
 - `replaceDuplicatePlugins`: Defaults to `false`. If `true`, the plugins in custom webpack config will replace the corresponding plugins in default Angular CLI webpack configuration. If `false`, the [default behavior](#merging-plugins-configuration) will be applied.
   **Note that if `true`, this option will override `mergeRules` for `plugins` field.**
 
-Webpack configuration can be also written in TypeScript. Given the following example:
+Webpack configuration can be also written in TypeScript. In this case, it is the appliaction's `tsConfig` file which will be use by `tsnode` for `customWebpackConfig.ts` execution. Given the following example:
 
 ```ts
 // extra-webpack.config.ts
-import * as webpack from 'webpack';
+import { Configuration } from 'webpack';
 
 export default {
   output: {
     library: 'shop',
     libraryTarget: 'umd',
   },
-} as webpack.Configuration;
+} as Configuration;
 ```
 
 Do not forget to specify the correct path to this file:
@@ -421,6 +421,7 @@ module.exports = async config => {
 Since Angular 8 `index.html` is not generated as part of the Webpack build. If you want to modify your `index.html` you should use `indexTransform` option.  
 `indexTransform` is a path (relative to workspace root) to a `.js` or `.ts` file that exports transformation function for `index.html`.  
 Function signature is as following:
+If `indexTransform` is writes in TypeScript, it is the application's `tsConfig` file which will be use by `tsnode` for `indexTransform.ts` execution.
 
 ```typescript
 (options: TargetOptions, indexHtmlContent: string) => string|Promise<string>;
--- a/packages/custom-webpack/examples/full-cycle-app/e2e/protractor.conf.js
+++ b/packages/custom-webpack/examples/full-cycle-app/e2e/protractor.conf.js
@@ -19,12 +19,20 @@ exports.config = {
   jasmineNodeOpts: {
     showColors: true,
     defaultTimeoutInterval: 30000,
-    print: function() {},
+    print: function () {},
   },
   onPrepare() {
+    const tsConfig = require('path').join(__dirname, './tsconfig.json');
     require('ts-node').register({
-      project: require('path').join(__dirname, './tsconfig.json'),
+      project: tsConfig,
     });
+
+    // Register paths in tsConfig
+    const tsconfigPaths = require('tsconfig-paths');
+    const { absoluteBaseUrl: baseUrl, paths } = tsconfigPaths.loadConfig(tsConfig);
+    if (baseUrl && paths) {
+      tsconfigPaths.register({ baseUrl, paths });
+    }
     jasmine.getEnv().addReporter(new SpecReporter({ spec: { displayStacktrace: true } }));
   },
 };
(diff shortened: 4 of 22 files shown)


just-jeb:
Hi. First of all thanks for your contribution - it is highly appreciated!  
Your PR in some way intersects with #824, and the reason the build is failing is the same reason.  
If you have time and desire to investigate this issue, it will be incredibly helpful.

just-jeb:
Hey, thanks for the contribution.  
Currently two things are missing here - updating e2e tests and docs.  
I'd start with e2e, since I have a strong suspicion they will start failing once you try to use `.ts` config with Karma builder like [here](https://github.com/just-jeb/angular-builders/pull/824/files#diff-a2d81be02f314726924854ce8ad329d2be1bb7f6d8fcc44f9cd9ad4d3c3f5963R103).

scripts/run-ci.sh
```diff
@@ -46,6 +46,10 @@ publishToLocalRegistry
 
 yarn bootstrap:examples
 
+# Get travis's chrome version and download the appropriate webdriver-manager for protractor
+CHROME_VERSION=`google-chrome --version | egrep -o '[0-9.]+' | head -1`
+npx lerna exec --ignore '@angular-builders/*' -- ./node_modules/protractor/bin/webdriver-manager update --versions.chrome $CHROME_VERSION
```
muuvmuuv, line 51: What is the reason for this?

packages/custom-webpack/package.json
```diff
@@ -42,6 +42,7 @@
     "@angular-devkit/core": "^11.0.0",
     "lodash": "^4.17.15",
     "ts-node": "^9.0.0",
+    "tsconfig-paths": "^3.9.0",
```
muuvmuuv, line 45: I am currently merging part of your changes into my PR (#824) and adding `tsconfig-paths` was not needed. Why are you adding it, are there any benefits?

just-jeb: As far as I remember `ts-node` does not support `paths` property, so if you want to have the exact same config (including paths) for your webpack (or index) configuration you need this one.

muuvmuuv: Makes sense! But `ts-node` isn't needed for any Angular project since it compiles through `tsc` and does not require to be "started" through Node, so I wouldn't add this or do I miss any usecase here?

just-jeb: `ts-node` is indeed not needed for Angular app, because Angular compile (not exactly `tsc`) builds it, that's correct.  
However, index transform and webpack config are not part of the app, they are part of the build chain, which in our case is `custom-webpack` builder. The builders needs `ts-node` in order to understand the TS syntax of the files it loads (webpack config and index transform).

scripts/run-ci.sh
michelherv: I cherry picked this from #886. It was to skip resolved errors. Now that #886 is merged, I rebase this pull request onto master and remove cherry picked commits
Thanks for the question

michelherv:
Hi,
I updated e2e tests. Can you read the pull request again and tell me if tests are ok for you. Else can you help me by pushing some tests? I will try to improve the solution by TDD.

About the documentation, I have no improvement idear. `tsConfig` is a `@angular-devkit/architect:browser` property. It's not like `customWebpackConfig` or `indexTransform` which are custom properties. I don't know what to add about.

just-jeb:
@muuvmuuv @michelherv I appreciate the effort and thank you both, but currently we have two PRs that do practically the same. Can we please have a single PR that takes the best of the existing PRs?  
Would be much easier for me to follow and comment.  
Thank you!

muuvmuuv:
@michelherv @just-jeb before new year I won't have time for any OSS, so I would close mine and Michel can include all parts you both think that are missing here. 

Below or things I would change or add, doesn't mean they should be included in this PR.

- workspace npm scripts should use binaries directly instead of npx: https://github.com/just-jeb/angular-builders/pull/824/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R15
- puppeteer for karma testing browser bin to avoid using none-existing local binary: https://github.com/just-jeb/angular-builders/pull/824/files#diff-5948cfd5023315c8bda008c2c6688e5a232dd1e0e4932ccfa01db65743069077R3
- use karma-coverage: https://github.com/just-jeb/angular-builders/pull/824/files#diff-5948cfd5023315c8bda008c2c6688e5a232dd1e0e4932ccfa01db65743069077R18
- just use the version without `&copy;` for better test parsing: https://github.com/just-jeb/angular-builders/pull/824/files#diff-f91984e0f1388cbc9fd6c3fb6ba9dbe36b6b08425104d2af479e64ef4b38dfd7R25
- add some note why we use the `node`-typing: https://github.com/just-jeb/angular-builders/pull/824/files#diff-a5e05c6f6ee63c72a42f6f6662bc73589c4c7880a083f414c6441267d53694ebR6
- use file import: https://github.com/just-jeb/angular-builders/pull/824/files#diff-8befd2cce9bc793e8dc30fc050dbbf77f613ef94042f12fb1dca6f1e3b746fc1R8
- fix target option typing bug (appeared in my installation): https://github.com/just-jeb/angular-builders/pull/824/files#diff-b2c00b61854bfb83318b9fa5f3c5110bae9a2fba71f2e9a973059665f6d0651cR17

just-jeb:
Hey, sorry it takes me so long to respond, I'm trying to make the `v11` stable (#908).  
> Hi,
> I updated e2e tests. Can you read the pull request again and tell me if tests are ok for you. Else can you help me by pushing some tests? I will try to improve the solution by TDD.
> 
> About the documentation, I have no improvement idear. `tsConfig` is a `@angular-devkit/architect:browser` property. It's not like `customWebpackConfig` or `indexTransform` which are custom properties. I don't know what to add about.

I'll take a look at tests and leave my comments as a review.  
Regarding documentation; by adding docs I meant that we need to specify in the docs that `indexTransform` can now be written with Typescript, and that both `customWebpackConfig.ts` and `indexTransfrom.ts` are using the app tsConfig now.  
Also if any changes have to be applied to `tsconfig.json` in order to make it work, (like `"types": [ "node" ]`) then we should mention it in the docs as well.

.github/workflows/angular-builders.yml
```diff
@@ -18,12 +18,12 @@ jobs:
         with:
           token: ${{ secrets.GH_PERSONAL_TOKEN }}
 
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
```
just-jeb, line 21: I think it's better to keep it as a separate PR, in order to avoid mixing features.

packages/custom-webpack/examples/full-cycle-app/angular.json
```diff
@@ -18,9 +18,12 @@
           "builder": "@angular-builders/custom-webpack:browser",
           "options": {
             "customWebpackConfig": {
-              "path": "./extra-webpack.config.ts"
+              "path": "extra-webpack.config.ts",
+              "mergeStrategies": {
```
just-jeb, line 22: [`mergeRules`](https://github.com/just-jeb/angular-builders/blob/master/MIGRATION.MD#migration-from-version-10-to-version-11) should be used instead

packages/custom-webpack/examples/full-cycle-app/tsconfig.app.json
```diff
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./out-tsc/app",
-    "types": []
+    "types": [ "node" ]
```
just-jeb, line 5: I understand that this is necessary for `ts-node` to work? If yes, maybe we can put it into the hardcoded `compileOptions` when registering the `ts-node`?  
Alternatively the docs should mention that one has to add this to the config when working with `.ts` versions of `customWebpackConfig` or `indexTransform`.

just-jeb (commented):
Looks good overall. But I think we also need a test that makes sure `indexTransform` uses the app config as well.

michelherv:
Hi,
I have a problem with the build process
image

Do you understand what is the problem ?

just-jeb:
> Hi,
> I have a problem with the build process
> image
> 
> Do you understand what is the problem ?

Oh, I think I read something about it. As far as I remember it's because it's a fork, and you don't have a token defined in your GH secrets.   
Let me try to find the issue.

just-jeb:
Here it is: https://github.com/actions/checkout/issues/298

just-jeb:
Seems like there is a workaround. I'll try to apply it.

just-jeb:
Aaaand fixed!

just-jeb:
@michelherv are we good to go? Can this be merged?

michelherv:
Hi @just-jeb , it is ok now for merging. Sorry for the late, I was very busy these last days.

According @muuvmuuv [suggestions](https://github.com/just-jeb/angular-builders/pull/879#issuecomment-744495019) :
- I just use the version without &copy; for better test parsing ;
- I add some note why we use the node-typing.

The others suggestions are also very important. I think we should have to treat them as real independent improvements which will be keep anyway independently of this PR. They would can be merge before or after this PR. It it why I opened the PR #925

just-jeb (approved):

packages/custom-webpack/examples/full-cycle-app/angular.json
```diff
@@ -50,7 +53,6 @@
               "optimization": true,
               "outputHashing": "all",
               "sourceMap": false,
-              "extractCss": true,
```
tony, line 53: Any reason why this was removed?

tony: (re: https://github.com/just-jeb/angular-builders/issues/944#issuecomment-863207093)

tony, line 53: > Option `"extractCss"` is deprecated: Deprecated since version 11.0. No longer required to disable CSS extraction for HMR.

PR #941, opened 2021-02-28
fix(custom-webpack): change `ts-node` compilation target to include TS features

arturovt:
Hey @just-jeb . I've removed `target` property that was introduced in this PR (https://github.com/just-jeb/angular-builders/commit/c85cd103af6047fefbea273ace1d9446829f3651) since it's actually not needed. This property didn't exist in `@10` version and it worked perfect, but we've upgraded to `@11` version and started getting such errors.

This property will require maintaining since it will require bumping target each time TS releases new features, which brings additional complexity.

image

---

And with a new change:

image


tony:
@arturovt Could this be related to the issue at #944?

I get an unexpected token as well, though it's in a SCSS file

arturovt:
Hey @tony . Not sure about this. I would have looked at the reproduction but I've seen there is a huge repo (the repo where your build is failing).

just-jeb (approved):

PR #971, opened 2021-04-19
fix(custom-webpack): specify possible types of customWebpackConfig values

juristr:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

At the moment if someone uses this with the latest versions of [Nx](https://nx.dev) and does not specify a `webpack.config.js` the build fails because it requires one.

The reason is how the defaults are being interpreted by the Angular CLI vs Nx schema parser. Right now the `custom-webpack` browser schema defines `customWebpackConfig` to be of type `object` but assigns a boolean value as the default. The Angular CLI is less strict in that it accepts that, while Nx is more strict (I'm working on the Nx side to avoid such edge cases as much as possible).

## What is the new behavior?

This PR adjusts the schema to properly reflect the possible types, basically a `object` or `boolean`. This is achieved with the `oneOf` property in the schema file.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
+++ b/packages/custom-webpack/e2e/custom-webpack-config-schema.ts
@@ -1,19 +1,29 @@
 export const customWebpackConfig = {
-  type: 'object',
   description: 'Custom webpack configuration',
-  properties: {
-    path: { type: 'string', description: 'Path to the custom webpack configuration file' },
-    mergeRules: {
-      type: 'object',
-      description:
-        'Merge rules as described here: https://github.com/survivejs/webpack-merge#mergewithrules',
-    },
-    replaceDuplicatePlugins: {
-      type: 'boolean',
-      description: 'Flag that indicates whether to replace duplicate webpack plugins or not',
-    },
-  },
   default: false,
+  oneOf: [
+    {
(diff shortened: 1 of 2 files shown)


just-jeb:
I think the default value `false` is a leftover from the very early days of the builder.  
The right type would be then `object` or `undefined` with no default value (this should fix the problem with `Nx` as well).

juristr:
@just-jeb passing undefined/null or no default value wouldn't work as the Angular CLI (and Nx) initialize those values with a default object.

image

Thus the check in the `CustomWebpackBuilder` fails and would require the user to specify a custom webpack configuration.

I don't think using `false` as the default value is wrong at all, it's just the schema that needs to be adjusted to account for that. Very similar to what the Angular devkit builder does for the `optimization` property. Let me know what you think

juristr:
@just-jeb Sorry for pinging 😅. Any opinion on the above?

just-jeb:
Sounds good to me

just-jeb (approved):

just-jeb:
@all-contributors please add @juristr for code

Issue #1031, opened 2021-09-03
No implicit any tsconfig cause issue with webpack config factory


GerkinDev:
## Describe the Bug

I'm using a factory to generate my webpack browser & server configuration. My application tsconfig is in full strict mode. When running the builder for the browser, it seems that type annotations are removed **before** typescript compilation, causing errors related to implicit `any` types.

## Minimal Reproduction

```ts
import { Configuration } from 'webpack';

export type ConfigFactory = ( baseConfig: Configuration, ngBuildConfig: any, ngBuildTarget: any ) => Configuration;

const factory: ConfigFactory = ( baseConfig, ngBuildConfig, ngBuildTarget ) => {
    // ...
}
export default factory;
```

Build fails with following message:

```
An unhandled exception occurred: ⨯ Unable to compile TypeScript:
configs/build/webpack.browser.config.ts:15:18 - error TS7006: Parameter 'baseConfig' implicitly has an 'any' type.

15 const factory = (baseConfig, ngBuildConfig, ngBuildTarget) => {
                    ~~~~~~~~~~
configs/build/webpack.browser.config.ts:15:30 - error TS7006: Parameter 'ngBuildConfig' implicitly has an 'any' type.

15 const factory = (baseConfig, ngBuildConfig, ngBuildTarget) => {
                                ~~~~~~~~~~~~~
configs/build/webpack.browser.config.ts:15:45 - error TS7006: Parameter 'ngBuildTarget' implicitly has an 'any' type.

15 const factory = (baseConfig, ngBuildConfig, ngBuildTarget) => {
                                               ~~~~~~~~~~~~~

See "/tmp/ng-lUATx4/angular-errors.log" for further details.
```

## Expected Behavior

Build should work.

## Environment

<pre><code>
Libs
- @angular/core version: 12.0.5
- @angular-devkit/build-angular version: 12.2.4
- @angular-builders/custom-webpack version: 12.1.1

For Tooling issues:
- Node version: 16.7.0
- Platform:  Fedora Linux x64

Others:

</code></pre>

## Additional Context

Seemms related to #749.

I've managed to fix the issue localy by editing https://github.com/just-jeb/angular-builders/blob/66b21aa50d31ae06ff4a3467851f068680f70cc9/packages/custom-webpack/src/utils.ts#L9-L17 and adding `transpileOnly`:
```ts
    require('ts-node').register({
      project: tsConfig,
      compilerOptions: {
        module: 'CommonJS',
[...]

PR #1035, opened 2021-09-05
Fix/1031

GerkinDev:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1031 

## What is the new behavior?

Ensures ts-node register is called once, and if tried to be called multiple times, asserts that all uses same tsconfig.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/custom-webpack/src/utils.ts
+++ b/packages/custom-webpack/src/utils.ts
@@ -1,11 +1,16 @@
-/**
- * check for TS node registration
- * @param file: file name or file directory are allowed
- * @todo tsNodeRegistration: require ts-node if file extension is TypeScript
- */
-export function tsNodeRegister(file: string = '', tsConfig?: string) {
-  if (file && file.endsWith('.ts')) {
-    // Register TS compiler lazily
+const _tsNodeRegister = (() => {
+  let lastTsConfig: string | null | undefined;
+  return (tsConfig?: string) => {
+    // Check if the function was previously called with the same tsconfig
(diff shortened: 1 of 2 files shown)


packages/custom-webpack/src/utils.ts
```diff
@@ -1,11 +1,16 @@
...
- * @todo tsNodeRegistration: require ts-node if file extension is TypeScript
- */
-export function tsNodeRegister(file: string = '', tsConfig?: string) {
-  if (file && file.endsWith('.ts')) {
-    // Register TS compiler lazily
+const _tsNodeRegister = (() => {
+  let lastTsConfig: string | null | undefined;
+  return (tsConfig?: string) => {
+    // Check if the function was previously called with the same tsconfig
+    if (typeof lastTsConfig !== 'undefined' && (tsConfig ?? null) !== lastTsConfig) {
```
just-jeb, line 5: I think you can simplify it even more by initializing the `lastTsConfig` to `null`. Thus the only check you'll have to do is the equality check.

just-jeb (approved):

packages/custom-webpack/src/utils.ts
GerkinDev: I wasn't sure if you always call this function with a tsconfig parameter, since it is marked as optional in the calling function. So, to cover all cases, I did it this way. If the tsconfig parameter is always set, this simplification would still cover all cases, but the call signature should be changed.

PR #1041, opened 2021-09-29
fix(custom-webpack): warn about multiple tsconfigs for custom config

### LINKED ISSUE #1037: v12.1.2 breaks SSR compilation with multiple tsconfigs.
## Describe the Bug

On `v12.1.2`, my SSR application fails to build when running `npm run dev:ssr` with the error: 
`Called with multiple tsconfigs.`

The error is not present in `v12.1.1` and the application compiles as expected.

I've tested downgrading from `v12.1.2` to `v12.1.1` on two projects and can confirm the issue is present in both projects.

The error seems to be thrown from [here](https://github.com/just-jeb/angular-builders/blob/359a4c4c6200bbe522516c9c7eda30614b2ede3c/packages/custom-webpack/src/utils.ts#L6)

## Minimal Reproduction

1) Install `v12.1.2` in a project using `@angular-builders/custom-webpack:server` to compile the server and `@angular-builders/custom-webpack:browser` for the browser application.
2) Attempt to build the project and receive the error above.
3) Downgrade to `v12.1.1`.
4) Attempt to build the project (which should succeed)

## Expected Behavior

The Angular application should compile both the frontend and server portions of the project correctly without errors.

## Screenshots

Screenshot

## Environment

<pre><code>
Libs
- @angular/core version: ^12.2.6
- @angular-devkit/build-angular version: ^12.2.6
- @angular-builders/custom-webpack: 12.1.2

For Tooling issues:
- Node version: v14.15.4
- Platform: Linux

Others:
- PNPM version: 6.13.0
</code></pre>

just-jeb:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1037 

## What is the new behavior?

Multiple `ts-node` registration attempts result in a warning, only first is really registered.

```
[ ] Yes
[x] No
```

## Other information

Initially issue was introduced with PR #1035. The original PR aimed to solve `ts-node` double registration with the same tsconfig and the base assumption was that it would never be called with two different tsconfigs.  
That assumption, however, turned to be wrong - in #1037 we see that the usage of `@nguniversal/builders` results in `browser` and `server` builders run in the same process, which effectively throws an exception.  
This PR fixes this use case and presents a warning instead of exception.  

@GerkinDev FYI.

### DIFF
--- a/packages/custom-webpack/src/custom-webpack-builder.spec.ts
+++ b/packages/custom-webpack/src/custom-webpack-builder.spec.ts
@@ -71,7 +71,8 @@ describe('CustomWebpackBuilder', () => {
       null,
       baseWebpackConfig,
       {},
-      targetOptions
+      targetOptions,
+      {} as any
     );
 
     expect(mergedConfig).toEqual(baseWebpackConfig);
@@ -85,7 +86,8 @@ describe('CustomWebpackBuilder', () => {
       {},
       baseWebpackConfig,
       {},
-      targetOptions
+      targetOptions,
+      {} as any
     );
 
     try {
@@ -110,7 +112,8 @@ describe('CustomWebpackBuilder', () => {
       { path: fileName },
       baseWebpackConfig,
       {},
-      targetOptions
+      targetOptions,
+      {} as any
     );
 
     try {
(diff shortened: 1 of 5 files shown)


GerkinDev:
Thank you. Sorry for the thrown error: at least you know that there are cases where conflicting configuration may happen.

just-jeb:
It's all good, I reviewed it, remember? 😄  
I also presumed it's not a valid use case, but apparently it is.

PR #1117, opened 2022-01-26
feat(custom-webpack): support ESM

### LINKED ISSUE #1101: Support ESM applications and configs
this package cannot be used with es modules (i.e packages that have `type: module`)

in [this line](https://github.com/just-jeb/angular-builders/blob/74f3bc99e15a78a9e1bf669b3cb553ed77756aa1/packages/custom-webpack/src/custom-webpack-builder.ts#L75) you are using `require()` to load the custom webpack config file. if custom.webpack.config.ts is inside an es module, you must use `import()`

```
// should be replaced with `import(path)`
  const customWebpackConfig = require(path);
```

also it has to handle webpack.config that written in typescript i.e `custom.webpack.ts`
you will need to pass `--experimental-specifier-resolution=node` to ts-node, so it can work with `.ts`

I use Angular v13 and custom-webpack v13

[minimal reproduction](https://stackblitz.com/edit/node-mdn6pg)
run `ng build`

just-jeb:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

A build with `@angular-builders/custom-webpack` builder will fail if ran inside ESM app

## What is the new behavior?

ESM fully supported

```
[ ] Yes
[x] No
```
## Other information

Closes #1101


just-jeb:
@eng-dibo I'm not sure if we should add something specific about ESM support in the docs. WDYT?

eng-dibo:
> add something specific about ESM support in the docs

Sure, it is a good feature and trendy these days. I see many mega packages updated their docs to mention that it now supports esm.

I see you merged the PR but the version in npm still 13.0.0

eng-dibo:
as you mentioned in the updated docs, tsconfig-paths doesn't support esm, alternatively there is another one tsc-alias that supports esm and runs in compile time instead of runtime (that gives better performance, but doesn't support runtime modules such as ts-node. I already opened an issue there)

usage:
`tsc & tsc-alias`

PR #1307, opened 2022-11-01
feat(custom-webpack): provide `verbose` option

arturovt:
This is a follow-up PR of the #633, since it's never been updated after and I just forgot about it. I'l mark it as a draft for now just to wait for your comments.

### DIFF
--- a/packages/custom-webpack/src/custom-webpack-builder.ts
+++ b/packages/custom-webpack/src/custom-webpack-builder.ts
@@ -1,4 +1,6 @@
+import { inspect } from 'util';
 import { getSystemPath, logging, Path } from '@angular-devkit/core';
+import { get } from 'lodash';
 import { Configuration } from 'webpack';
 import { CustomWebpackBrowserSchema } from './browser';
 import { CustomWebpackBuilderConfig } from './custom-webpack-builder-config';
@@ -25,7 +27,7 @@ type CustomWebpackConfig =
 export class CustomWebpackBuilder {
   static async buildWebpackConfig(
     root: Path,
-    config: CustomWebpackBuilderConfig,
+    config: CustomWebpackBuilderConfig | null,
     baseWebpackConfig: Configuration,
     buildOptions: any,
     targetOptions: TargetOptions,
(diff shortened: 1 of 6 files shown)


packages/custom-webpack/src/custom-webpack-builder.ts
```diff
@@ -73,3 +81,22 @@ async function resolveCustomWebpackConfig(
...
+): void {
+  // There's no reason to log the entire configuration object
+  // since Angular's Webpack configuration is huge by default
+  // and doesn't bring any meaningful context by being printed
+  // entirely. Users can provide a list of properties they want to be logged.
+  if (config.verbose?.properties) {
+    for (const property of config.verbose.properties) {
+      const value = get(webpackConfig, property);
+      if (value) {
+        logger.info(value);
```
just-jeb, line 98: I'm not sure it will print the whole thing properly. I mean yeah, if the property is a primitive then sure, but what if I want to print out, say, `loaders`? Or `plugins`, especially if one of them is a class instance with a constructor. I don't think that `logger.info` knows how to deal with it. That's the challenge :)

just-jeb:
If/when this is merged, this is gonna be so useful! We could even include it as a requirement in issue template - run the build with `--verbose` and add it to the issue description. Can't wait to see it working ❤️.

packages/custom-webpack/src/custom-webpack-builder.ts
arturovt: Yeah that's true. Lemme just check how their logger prints objects. I'll respond in the chat. We prolly have 3 ways to go with:
- use their logger
- use 3rd-party logger
- serialize the object into a string and log as a string (meh)

just-jeb: I think we should stick with their logger, so probably the `meh` option is the way to go 😅.  
Whether to use some 3rd party for serialization or not is another question, but as for logger I believe it would be best to continue using their logger.

just-jeb (approved):

PR #1325, opened 2022-11-29
feat(jest): support Jest config in different formats

### LINKED ISSUE #1057: @angular-builders/jest - Support jest.config in typescript
According https://jestjs.io/docs/configuration it's possible to use a Typescript jest-config if `ts-node` is installed, but I tried it and I got
`An unhandled exception occurred: Cannot use import statement outside a module`

Support a typescript option for jest config file.

**Describe alternatives you've considered**
I'm trying to customize the `jsdom` options and I just cannot do it with plain javascript it seems:
```
const VirtualConsole = new require('jsdom/lib/jsdom/virtual-console');

const virtualConsole = (new VirtualConsole()).sendTo(console, { omitJSDOMErrors: true });
console.log(virtualConsole)

process.env.TZ = 'UTC';

module.exports = {
  testRunner: 'jest-jasmine2',
  testEnvironment: 'jsdom',
  testEnvironmentOptions: { virtualConsole },
```

arturovt:
Closes #1057

### DIFF
--- a/packages/jest/src/types.ts
+++ b/packages/jest/src/types.ts
@@ -1,0 +1,3 @@
+// The type is used to replace the `any` type to be more explicit with return types.
+// We can't use the `Config` type from `jest` since this might be a breaking change.
+export type JestConfig = Record<string, any>;
--- a/packages/jest/src/types.ts
+++ b/packages/jest/src/types.ts
@@ -1,3 +1,1 @@
-// The type is used to replace the `any` type to be more explicit with return types.
-// We can't use the `Config` type from `jest` since this might be a breaking change.
-export type JestConfig = Record<string, any>;
+export type JestConfig = import('jest').Config;
--- a/examples/jest/simple-app/angular.json
+++ b/examples/jest/simple-app/angular.json
@@ -90,6 +90,18 @@
           "builder": "@angular-builders/jest:run",
           "options": {}
         },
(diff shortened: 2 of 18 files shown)


packages/jest/src/types.ts
```diff
@@ -0,0 +1,3 @@
+// The type is used to replace the `any` type to be more explicit with return types.
+// We can't use the `Config` type from `jest` since this might be a breaking change.
```
just-jeb, line 2: Why would it be a breaking change?

just-jeb (approved):

packages/jest/src/types.ts
arturovt: It doesn’t ensure to be a breaking change. I’m just _afraid_ it to be a breaking change.

arturovt:
I've updated the type to point to `jest.Config`.

packages/jest/src/types.ts
just-jeb: Yeah I understand, I'm just not sure why 😁. Do you have a specific scenario in mind?

arturovt: I just thought initially that someone may `require` this package and might have issues with typings.

arturovt: But I ain't sure if someone will ever `require` it.

Issue #1408, opened 2023-06-21
tsConfig option doesn't work


marek-siemieniuk-morawski:
## Describe the Bug

When I have a workspace with multiple projects:

```
{
  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
  "version": 1,
  "newProjectRoot": "projects",
  "projects": {
    "foo": {
      "projectType": "application",
      "schematics": {},
      "root": "projects/foo",
      "sourceRoot": "projects/foo/src",
      "prefix": "app",
      "architect": {
        "build": {},
        "serve": {},
        "extract-i18n": {},
        "test": {
          "builder": "@angular-builders/jest:run",
          "options": {
            "polyfills": [
              "zone.js",
              "zone.js/testing"
            ],
            "assets": [
              "projects/foo/src/favicon.ico",
              "projects/foo/src/assets"
            ],
            "styles": [
              "projects/foo/src/styles.css"
            ],
            "scripts": []
          }
        }
      }
    },
    "boo": {}
  }
}
```

the expected `tsconfig.spec.json` location is in the workspace root, not in the project directory (`projects/foo`):

```
> angular-jest@0.0.0 test:foo
> ng run foo:test

 FAIL  projects/foo/src/app/app.component.spec.ts
  ● Test suite failed to run

    File not found: <rootDir>/tsconfig.spec.json (resolved as: /Users/marek/Code/angular-jest/tsconfig.spec.json)

      at NgJestConfig.ConfigSet.resolvePath (node_modules/ts-jest/dist/legacy/config/config-set.js:598:19)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        0.941 s
Ran all test suites.
```

Same, when I set the `tsConfig` option to set a custom path to `tsconfig.spec.json`, it throws the same error:

```
> angular-jest@0.0.0 test:foo
> ng run foo:test

 FAIL  projects/foo/src/app/app.component.spec.ts
  ● Test suite failed to run

    File not found: <rootDir>/tsconfig.spec.json (resolved as: /Users/marek/Code/angular-jest/tsconfig.spec.json)

      at NgJestConfig.ConfigSet.resolvePath (node_modules/ts-jest/dist/legacy/config/config-set.js:598:19)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        0.941 s
Ran all test suites.
```

## Minimal Reproduction

[...]

PR #1443, opened 2023-08-14
fix(jest): set path to tsconfig correctly

pixtron:
## PR Checklist

## PR Type
- [x] Bugfix

Option `tsConfig` in angular.json is ignored. `tsconfig.spec.json` [seems to be loaded from the CWD](https://github.com/just-jeb/angular-builders/issues/1408#issuecomment-1609794674) instead of the project root.

Issue Number: #1408 

Eventually yes, if suddenly the path to tsconfig is configured correctly.

## Other information
Followed the docs of `jest-preset-angular` to override/extend the transform config.
https://thymikee.github.io/jest-preset-angular/docs/getting-started/presets/#advanced

Tested it locally by building and copying over the dist folder to node_modules/@angular-builders/jest.

### DIFF
--- a/examples/jest/multiple-apps/tsconfig.spec.json
+++ b/examples/jest/multiple-apps/tsconfig.spec.json
@@ -1,7 +1,0 @@
-{
-  "extends": "./tsconfig.base.json",
-  "compilerOptions": {
-    "outDir": "../../out-tsc/spec"
-  },
-  "include": ["**/*.spec.ts", "**/*.d.ts"]
-}
--- a/packages/jest/src/default-config.resolver.spec.ts
+++ b/packages/jest/src/default-config.resolver.spec.ts
@@ -10,7 +10,7 @@ describe('Resolve project default configuration', () => {
   it('Should resolve tsconfig relatively to project root', () => {
     const config = defaultConfigResolver.resolveForProject(normalize('/some/cool/directory'));
     expect(config.transform[defaultConfigResolver.tsJestTransformRegExp]).toEqual([
-      'ts-jest',
+      'jest-preset-angular',
       {
         tsconfig: getSystemPath(normalize(`/some/cool/directory/${tsConfigName}`)),
       },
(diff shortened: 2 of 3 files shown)


pixtron:
@just-jeb `defaultTransformerOptions` from `jest-preset-angular` where undefined. As i did not find a way to properly import them i just added that [one line directly](https://github.com/just-jeb/angular-builders/pull/1443/commits/51eb6423ee07faaf3e40fa15ffc3f236fd58b341#diff-93b4c0d79d72878fda7c6abc4885cbc0908746e583244f6d2e491ab5a012cd29R45).

After this change i can successfully run `yarn ci` in `packages/jest`. But when i run `yarn ci` for the whole repo i still get some error on [line 60 in `./scripts/run-ci.sh`](https://github.com/just-jeb/angular-builders/blob/8f00fd9b324d0151f722a4e962fcbeab7891baae/scripts/run-ci.sh#L60)

Tried to run `yarn ci` in all packages, but ``

just-jeb:
Hey, @pixtron, to make sure it really works I'd suggest you add an e2e test.  
In fact you don't even need to add a new one, just change one small thing, as I mentioned [here](https://github.com/just-jeb/angular-builders/issues/1408#issuecomment-1609890701), specifically: 

> That said, looks like our [multi-app example](https://github.com/just-jeb/angular-builders/tree/master/examples/jest/multiple-apps) suffers from the same issue, it's just never popped up because it has `tsconfig.spec.json` on the root level as well.
When I delete it all the tests of this example start to fail with the same error.

So need to delete `tsconfig.spec.json` from the root, make sure the tests fail (without your change), and then add your change and make sure the tests pass.

pixtron:
> So need to delete `tsconfig.spec.json` from the root, make sure the tests fail (without your change), and then add your change and make sure the tests pass.

Removed `examples/jest/multiple-apps/tsconfig.spec.json`. `yarn ci` passes locally. When the same change is applied to master `yarn ci` fails (as expected) with: 

```
+ set +x
+ yarn test my-first-app
+ tee tests.log
    warning: unable to locate custom jest configuration file at path "/examples/jest/multiple-apps/projects/my-first-app/jest.config.js"
FAIL projects/my-first-app/src/app/app.component.spec.ts
  ● Test suite failed to run

    File not found: <rootDir>/tsconfig.spec.json (resolved as: /examples/jest/multiple-apps/tsconfig.spec.json)
```

PR #1536, opened 2023-11-15
feat(custom-esbuild): allow providing ESBuild plugins

arturovt:


arturovt:
@just-jeb you can start reviewing the PR.

packages/custom-esbuild/src/application/schema.ext.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$id": "BuildCustomApplicationSchema",
+  "title": "Custom application schema for Build Facade",
+  "description": "Application target options"
```
just-jeb, line 4: I think we miss the actual extension here (`plugins`?)

packages/custom-esbuild/src/custom-esbuild-schema.ts
```diff
@@ -0,0 +1,9 @@
+import { ApplicationBuilderOptions, DevServerBuilderOptions } from '@angular-devkit/build-angular';
+
+export type CustomEsbuildApplicationSchema = ApplicationBuilderOptions & {
+  plugins?: string[];
+};
+
+export type CustomEsbuildDevServerSchema = DevServerBuilderOptions & {
+  middlewares?: string[];
```
just-jeb, line 8: If possible let's generate these with `quicktype` like [here](https://github.com/just-jeb/angular-builders/blob/master/packages/timestamp/package.json#L29), so that we have `.json` as a single source of truth for the schema.

packages/custom-esbuild/src/load-modules.ts
```diff
@@ -0,0 +1,25 @@
+import { Path, getSystemPath, logging } from '@angular-devkit/core';
```
just-jeb, line 1: I wonder if we should take this code into a common package (it becomes a pattern already, used in [`custom-webpack`](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/src/utils.ts), [`jest`](https://github.com/just-jeb/angular-builders/blob/master/packages/jest/src/utils.ts), here). WDYT? Looks like a good opportunity for a little refactoring.

just-jeb: If we decide to go with a `angular-builders/common`, then [`merge-schemes.ts`](https://github.com/just-jeb/angular-builders/blob/master/merge-schemes.ts) from the root can also go there.

packages/custom-esbuild/src/schema.ext.json
```diff
@@ -0,0 +1,12 @@
+{
```
just-jeb, line 1: This file is redundant since we have `application`? Or am I missing something?

packages/custom-esbuild/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "e2e": "jest --config ../../jest-e2e.config.js",
+    "clean": "rimraf dist",
+    "ci": "./scripts/ci.sh"
+  },
+  "builders": "builders.json",
+  "dependencies": {
+    "@angular-devkit/architect": ">=0.1700.0 < 0.1800.0",
+    "@angular-devkit/build-angular": "^17.0.0",
+    "@angular-devkit/core": "^17.0.0",
+    "esbuild": "0.19.5",
```
just-jeb, line 46: We only need it for types so it should be in `devDependencies` (in order to avoid version conflicts with Angular CLI)

examples/custom-esbuild/sanity-esbuild-app/angular.json
```diff
@@ -0,0 +1,126 @@
+{
+  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
```
just-jeb, line 2: Looks good, do you think we need to add another sanity app which is ESM by default (like [here](https://github.com/just-jeb/angular-builders/blob/master/examples/custom-webpack/sanity-app-esm/angular.json))?

examples/custom-esbuild/sanity-esbuild-app/esbuild.config.js
```diff
@@ -0,0 +1,13 @@
+const defineTextPlugin = {
```
just-jeb, line 1: What is the purpose of this file, where do we use it?

examples/custom-esbuild/sanity-esbuild-app/angular.json
```diff
@@ -0,0 +1,126 @@
...
+              "plugins": ["esbuild/define-text-plugin.mjs"]
+            },
+            "cjs": {
+              "plugins": ["esbuild/define-text-plugin.js"]
+            }
+          },
+          "defaultConfiguration": "production"
+        },
+        "serve": {
+          "builder": "@angular-builders/custom-esbuild:dev-server",
```
just-jeb, line 57: How are we testing custom middlewares?

just-jeb (changes requested):

muuvmuuv:
If you're going to do a beta release, I can test the new builder in our application?

just-jeb:
@muuvmuuv Yes, it will first go to a beta release and you're welcome to test it. Once this PR is merged a beta version will be created automatically. 
@arturovt are we waiting for more changes here?

arturovt:
@just-jeb yeah I resolved the comments what I made fixes for. I wasn't able to use `quicktype`, I don't get how it works.

I also have to get to the stuff @muuvmuuv mentioned in the issue.

arturovt:
@muuvmuuv I see the `extensions` change they made is not exposed yet and it's going to target 17.1.0.

arturovt:
> > @muuvmuuv I see the `extensions` change they made is not exposed yet and it's going to target 17.1.0.
> 
> It just got released 🎉 anything I can help here with?

I added the HTML transformer support.

muuvmuuv:
Just linked the PR locally and tested both builders against our application, where I used my custom cli script before. 
The application builders works perfectly! My plugins work and the transformer works, thanks.
Unfortunately, the dev-server won't work and will not until Angular takes action in this if-case here because it will never run Vite for custom builders. 
https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/dev-server/builder.ts#L65
I tested to override the `buildName` with `builderSelector` in the pipe where it returns `middleware, buildPlugins` but that would mean Vite will use the angular application builder not the custom one. So yeah... a blocker

muuvmuuv:
Great find! Excelent.

arturovt:
@dmitry-stepanenko thanks for pinpointing. Do you wanna participate in finalizing these changes? Because my mind is exhausting :smile: 

P.S. you can DM me in google chat.

muuvmuuv:
Also noticed that in `executeDevServerBuilder.transforms` the `indexHtml` is missing.

muuvmuuv:
Hey, I made a few more changes to your latest ones. It works on my project now, including index.html transformation.

[custom-esbuild-changes.diff.txt](https://github.com/just-jeb/angular-builders/files/14072254/custom-esbuild-changes.diff.txt)

arturovt:
> Hey, I made a few more changes to your latest ones. It works on my project now, including index.html transformation.
> 
> [custom-esbuild-changes.diff.txt](https://github.com/just-jeb/angular-builders/files/14072254/custom-esbuild-changes.diff.txt)

Nice, thanks! I pushed your changes too. I think we're on a final spot. @just-jeb this one can be reviewed, but the CI is failing by some reason, tho it doesn't fail locally :shrug:

packages/custom-esbuild/src/dev-server/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  options: CustomEsbuildDevServerSchema,
+  context: BuilderContext
+): Observable<DevServerBuilderOutput> {
+  const buildTarget = targetFromTargetString(
+    // `browserTarget` has been deprecated.
+    options.buildTarget ?? options.browserTarget!
+  );
+
+  async function getBuildTargetOptions() {
+    return (await context.getTargetOptions(
```
muuvmuuv, line 31: Guess this can be simplified now

just-jeb (commented):
The code looks good to me, except for a few comments (please address them). 
The only real blocker for merging it to master is the lack of README on a package level and a link to it on a repo level, otherwise we're good to go. 
@arturovt I fixed the conflicts, but the `sanity-esbuild-app` fails its test locally as well as in CI.

packages/custom-esbuild/src/dev-server/index.ts
muuvmuuv: getBuildTargetOptions is not needed as a separate function, getTargetOptions can be used directly in "from".

arturovt: Its return type is `type|Promise`. I used the separate function explicitly to await the value (and also to cast it).

packages/custom-esbuild/src/dev-server/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      const middleware = await Promise.all(
+        (options.middlewares || []).map(path =>
+          // https://github.com/angular/angular-cli/pull/26212/files#diff-a99020cbdb97d20b2bc686bcb64b31942107d56db06fd880171b0a86f7859e6eR52
+          loadModule<Connect.NextHandleFunction>(workspaceRoot, path, tsConfig, context.logger)
+        )
+      );
+
+      const buildPlugins = await Promise.all(
+        (buildOptions.plugins || []).map(path =>
+          loadModule<Plugin>(workspaceRoot, path, tsConfig, context.logger)
```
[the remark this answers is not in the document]
arturovt: I’ll work on it later after this one is merged.

arturovt:
Alright, I’ll try to run tests locally from my Mac and see what’s necessary to resolve them.

packages/custom-esbuild/src/dev-server/index.ts
muuvmuuv: OK, makes it more clean, ofc with casting. 
Didn't know there is a difference between returning the promise and awaiting it in a separate function (will put in on my reading list).

arturovt:
@just-jeb was able to resolve tests locally as thus the `yarn ci` for the `custom-esbuild` is green. Just can't see the GH results.

just-jeb (approved):

muuvmuuv:
My man! Thank you, already made my application zoneless and this will just boost it to space 🚀

just-jeb:
@muuvmuuv How is it going with beta? All good? Can we graduate?

muuvmuuv:
Been using it since you published it no issues yet build and dev works like a charm

PR #1654, opened 2024-01-31
refactor(common): move `loadModule` in shared package

arturovt:


packages/common/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@angular-builders/common",
+  "version": "0.0.0",
```
just-jeb, line 3: Let's make it 1.0.0?

just-jeb (approved):

Issue #1213, opened 2022-05-17
Failed to run typescript webpack-config.ts with custom webpack plugin in Angular 13 from compiled mjs format module


sergsar:
## Describe the Bug

Failed to run typescript webpack-config with custom webpack plugin in Angular 13 from compiled mjs module.

`An unhandled exception occurred: Unknown file extension ".ts" for /Users/.../webpack-plugin-use-test/webpack.config.ts`

`[error] TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/sergej/Projects/local/ng-documentation-test/webpack.config.ts
    at new NodeError (node:internal/errors:371:5)
    at Object.file: (node:internal/modules/esm/get_format:72:15)
    at defaultGetFormat (node:internal/modules/esm/get_format:85:38)
    at defaultLoad (node:internal/modules/esm/load:13:42)
    at ESMLoader.load (node:internal/modules/esm/loader:303:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:230:58)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:244:11)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)`

## Minimal Reproduction

1. Create new Angular 13 empty project
2. Compile any Webpack Plugin with Angular 13 ng build project
3. Install compiled module with plugin to initial project
4. Install @angular-builders/custom-webpack
5. Add customWebpackConfig configuration with typescript file *.ts

`import {CustomWebpackTestPlugin} from "test-webpack-plugin";

export default {
  plugins: [
    new CustomWebpackTestPlugin(),
  ],
};`
6. Run ng serve or ng build

Could been checked in this project:
https://github.com/sergsar/webpack-plugin-use-test

## Expected Behavior

Run typescript webpack-config**.ts** with custom webpack plugin in Angular 13 from  .mjs compliled format module

## Environment

<pre><code>
Libs
- @angular/core version: 13.3.0
- @angular-devkit/build-angular version: 13.3.5
- @angular-builders/custom-webpack version: 13.1.0

For Tooling issues:
- Node version: 16.13.0
- Platform:  Mac

robertmazzo:
@sergsar Hi, what was the final result here? I'm currently struggling with Angular 13 on this same issue.

[...]

PR #1659, opened 2024-02-04
fix(common): update TS module resolution flow

### LINKED ISSUE #1197: Build Warning with Angular Universal - Register ts-node again
I am getting this warning with Angular Universal:

register ts-node

```
Trying to register ts-node again with a different tsconfig - skipping the registration.
                   tsconfig 1: C:\web projects\fireblog.io/tsconfig.server.json
                   tsconfig 2: C:\web projects\fireblog.io/tsconfig.app.json
```

This started once I switched over to Webpack.  Here is my [angular.json](https://github.com/jdgamble555/fireblog.io/blob/master/angular.json).  

I believe it to be a bug in the compilation...

J

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When loading `.ts` config/plugin files, ts-node is registered once per process. If two projects use different tsconfig files, the second registration is silently skipped with a warning, meaning the wrong tsconfig may be used. Additionally, registering ts-node with `module: 'CommonJS'` while the user's tsconfig specifies `moduleResolution: 'bundler'` or `'Node16'` caused TS5095/TS5110 errors.

Closes #1197, #1213

## What is the new behavior?

- ts-node is registered and unregistered around each `.ts` file load (register → require → unregister), so each load uses the correct project tsconfig
- `moduleResolution: 'node'` is added to the ts-node compiler options override to avoid TS5095/TS5110 conflicts when users specify `moduleResolution: 'bundler'` or `'Node16'`; this is safe under `transpileOnly: true` (no type-check semantics depend on it)
- The `@angular-devkit/core` logger dependency is removed from `common`; the logger parameter is removed from `loadModule()`
- Added integration test fixture: ESM package import from TypeScript webpack config (`webpack-esm-plugin` local package with CJS + ESM exports)

```
[x] Yes
[ ] No
```

The `loadModule()` signature change (removing the `logger` parameter) is a breaking change for external consumers. The register/unregister-per-load model is also a behavior change: `service.enabled(false)` disables the require hook process-wide, so concurrent loads can race. To be shipped in the next major.

## Other information

Originally authored by @arturovt. Rebased and fixed to resolve CI failures (TS5095/TS5110 moduleResolution conflict, missing `webpack-esm-plugin` CJS entry, removed ESM hook that caused TypeError in Angular build context).

### DIFF
--- a/examples/custom-webpack/sanity-app-esm/custom-webpack.config.ts
+++ b/examples/custom-webpack/sanity-app-esm/custom-webpack.config.ts
@@ -1,7 +1,14 @@
 import type { Configuration } from 'webpack';
 
+import { WebpackEsmPlugin } from 'webpack-esm-plugin';
+
 export default async (cfg: Configuration) => {
   const { default: configFromEsm } = await import('./custom-webpack.config.js');
+
+  // This is used to ensure we fixed the following issue:
+  // https://github.com/just-jeb/angular-builders/issues/1213
+  cfg.plugins!.push(new WebpackEsmPlugin());
+
   // Do some stuff with config and configFromEsm
   return { ...cfg, ...configFromEsm };
 };
--- a/examples/custom-webpack/sanity-app-esm/package.json
+++ b/examples/custom-webpack/sanity-app-esm/package.json
@@ -44,6 +44,7 @@
     "karma-jasmine": "5.1.0",
     "karma-jasmine-html-reporter": "2.2.0",
     "puppeteer": "24.43.1",
-    "typescript": "5.9.3"
+    "typescript": "5.9.3",
+    "webpack-esm-plugin": "file:./webpack-esm-plugin"
   }
 }
--- a/examples/custom-webpack/sanity-app-esm/webpack-esm-plugin/package.json
+++ b/examples/custom-webpack/sanity-app-esm/webpack-esm-plugin/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "webpack-esm-plugin",
+  "version": "0.0.1",
+  "module": "./webpack-esm-plugin.mjs",
+  "typings": "./webpack-esm-plugin.d.ts",
+  "exports": {
+    "./package.json": {
+      "default": "./package.json"
+    },
+    ".": {
+      "types": "./webpack-esm-plugin.d.ts",
+      "node": "./webpack-esm-plugin.mjs",
+      "default": "./webpack-esm-plugin.mjs"
+    }
+  },
+  "sideEffects": false
+}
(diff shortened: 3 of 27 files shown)


arturovt:
Ugh that CI is so weird with random failures... doesn't even show any error...

arturovt:
@just-jeb could you run ci locally? Maybe you can see the actual error…

just-jeb:
@arturovt I'll try. Were you unable to run it locally?

just-jeb:
**Code review note for Jeb:**

The `loadModule()` signature change (removing the `logger` parameter) is exported from `@angular-builders/common`'s public index. The PR description correctly flags this as technically a breaking change for external consumers. Tagging it `breaking-change` for visibility — but given that `@angular-builders/common` is an internal utility package with no public API guarantees, merging before the next major is a valid judgment call.

The core logic change (register/unregister ts-node around each .ts load instead of once-per-process) is a significant behavioral shift. CI passes all 44 tests which covers the expected scenarios. The main open risk is edge cases not captured by the test suite (very unusual module-loading sequences).

Ready to merge pending Jeb's review of the register/unregister pattern and the breaking-change decision.

just-jeb:
Thanks a lot for this, @arturovt — and apologies it sat so long. The problem it targets (#1197 / #1213 — loading multiple `.ts` configs that need different tsconfigs in one process) is real and **both issues stay open**; we'll tackle them properly. But I'm going to close this PR rather than merge it, for a few reasons:

1. **It re-introduces a regression we just fixed.** The new `register-ts-project.ts` sets `moduleResolution: 'node'` in the ts-node compiler options. We deliberately removed that override in #2187 to fix #2025 — forcing `node` resolution breaks subpath package exports (e.g. `@angular/core/primitives/di`) for configs using `bundler`/`node16` resolution. The per-load timing doesn't change that: resolution mode determines whether the subpath resolves, regardless of how briefly the hook is active.

2. **The register/unregister-per-load model reopens a global-state hazard.** ts-node's `register()` / `service.enabled()` mutate process-global module loading. Toggling that on/off around every `.ts` load means interleaving breaks it — a `.ts` config that `require()`s another `.ts`, or two loads in flight, and one load's teardown disables transpilation while another is mid-`require`. The current "register once + warn on a different tsconfig" guard exists precisely to avoid this class of bug. (The PR's own test note — "two parallel loads must each see a working register-require-unregister window" — points at exactly this.)

3. Minor: it doesn't invalidate `require.cache`, so a second load with a different tsconfig returns the already-cached module; and dropping the `logger` parameter from `loadModule()` is a breaking change.

For a proper fix later (v22 cycle), the shape I'd want is a **scoped, non-global** compiler — `ts-node.create()` plus explicit transpile per config — instead of toggling the global `require` hook. That keeps multi-tsconfig support without re-opening the global-registration problems.

Really appreciate the effort here — the diagnosis was useful even though we're going a different direction. Tracking the underlying bugs in #1197 / #1213.

PR #1705, opened 2024-03-04
docs(custom-esbuild): align the HTML transformer signature

### LINKED ISSUE #1690: TargetOptions parameter is undefined with custom-esbuild
## Describe the Bug

Hi, 

I am currently working on upgrading an Angular 16 application to Angular 17 and I would like to use esbuild.
My application requires index.html transformation. This transformation relies on data from the targetOptions parameter.

With version 16, I was using `@angular-builders/custom-webpack` and it was working fine.
Unfortunately, with `@angular-builders/custom-esbuild`, the targetOptions are not present during the transformation. Instead, it's the content of the index file.
Yet the [documentation ](https://github.com/just-jeb/angular-builders/tree/master/packages/custom-esbuild#index-transform) indicates that targetOptions should be accessible.

## Minimal Reproduction

Just creating a new Angular 17 application (via ng new), and following the instruction of the [doc](https://github.com/just-jeb/angular-builders/tree/master/packages/custom-esbuild) : 
- Adding `@angular-builders/custom-esbuild` as a dev dependency
- Replacing `@angular-devkit/build-angular:application` with `@angular-builders/custom-esbuild:application` and `@angular-devkit/build-angular:dev-server` with `@angular-builders/custom-esbuild:dev-server` in angular.json
- Adding index-html-transformer.ts with the following code : 
```
import { Target } from '@angular-devkit/architect/src/input-schema';
import { json } from '@angular-devkit/core';

type TargetOptions = json.JsonObject & Target;

export default (targetOptions: TargetOptions, indexHtml: string) => {
  console.log('targetOptions', targetOptions);

  const i = indexHtml.indexOf('</body>');
  const config = `<p>Configuration: ${targetOptions.configuration}</p>`;
  return `${indexHtml.slice(0, i)}
            ${config}
            ${indexHtml.slice(i)}`;
};
```
- Adding indexHtmlTransformer in angular.json

Here is the result on ng build / ng serve : 

custom esbuild ko
 
## Expected Behavior

TargetOptions should be there. Or at least, documentation needs an update.

## Environment

<pre><code>
Libs
- @angular/core version: 17.2.0
- @angular-devkit/build-angular version: 17.2.1
- @angular-builders/custom-esbuild: 17.1.0

For Tooling issues:
- Node version: 20.11.1
- Platform: Windows

arturovt:
Closes #1690

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,8 @@
     "update:examples": "yarn workspaces foreach --exclude '@angular-builders/*' run update:example"
   },
   "devDependencies": {
-    "@commitlint/cli": "^20.0.0",
-    "@commitlint/config-conventional": "^20.0.0",
+    "@commitlint/cli": "^21.0.0",
+    "@commitlint/config-conventional": "^21.0.0",
     "@lerna-lite/cli": "^5.0.0",
     "@lerna-lite/publish": "^5.0.0",
     "@types/lodash": "^4.14.118",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3255,189 +3255,187 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@commitlint/cli@npm:^20.0.0":
-  version: 20.5.3
-  resolution: "@commitlint/cli@npm:20.5.3"
+"@commitlint/cli@npm:^21.0.0":
+  version: 21.0.1
+  resolution: "@commitlint/cli@npm:21.0.1"
   dependencies:
-    "@commitlint/format": ^20.5.0
-    "@commitlint/lint": ^20.5.3
-    "@commitlint/load": ^20.5.3
-    "@commitlint/read": ^20.5.0
-    "@commitlint/types": ^20.5.0
+    "@commitlint/format": ^21.0.1
+    "@commitlint/lint": ^21.0.1
+    "@commitlint/load": ^21.0.1
+    "@commitlint/read": ^21.0.1
+    "@commitlint/types": ^21.0.1
     tinyexec: ^1.0.0
-    yargs: ^17.0.0
+    yargs: ^18.0.0
   bin:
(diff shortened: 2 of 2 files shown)


just-jeb:
Not sure if it actually closes #1690 - it does align the docs with actual implementation but from what I read there is a request for adding `options` (similarly to `custom-webpack` builder).

arturovt:
We can align the docs for now and then I’ll think on implementing options as a feature.

just-jeb:
Sure thing, just saying that aligning the docs won't close #1690 🙂

arturovt:
However, how do we include target options as the first parameter when the HTML transformer is called by Angular with only the HTML content? We are not responsible for its invocation with only a single parameter.

shisukei:
@arturovt I'm not sure if this is ideal, but from what I've seen, at the post-bundle time, if there is an index transformation function specified, then this function is called with the index content as a parameter.

Considering that the context is available in buildCustomEsbuildApplication (or in executeCustomDevServerBuilder), wouldn't it be acceptable to wrap this in a custom function? For instance, if we take the current code of this [file](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-esbuild/src/application/index.ts), we would replace line 30 with something like this:
`
return { codePlugins, indexHtmlTransformer: (indexHtml) => indexHtmlTransformer(context, indexHtml) } as ApplicationBuilderExtensions;
`

Not sure, but it seems to be a similar approach with custom-webpack ([source code](https://github.com/just-jeb/angular-builders/blob/master/packages/custom-webpack/src/transform-factories.ts#L42)).

just-jeb:
@arturovt Exactly what @shisukei said, we can create a wrapper function and use the target params from the closure, just like we do in `custom-webpack`.

PR #1711, opened 2024-03-08
fix(custom-esbuild): add extract-i18n builder

kamilchlebek:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

We successfully switched from using webpack to esbuild thanks to `@angular-builders/custom-esbuild:application`.

However once we tried to extract i18n messages our `xlf` file had 0 items inside. 

It turned out that angular checks builder name to decide whether to use webpack/esbuild strategy - https://github.com/angular/angular-cli/blob/c2b8d84e360f0508e73f7c502516f898f067c03b/packages/angular_devkit/build_angular/src/builders/extract-i18n/builder.ts#L64-L65

## What is the new behavior?

Proposed `@angular-builders/custom-esbuild:extract-i18n` builder temporarily overrides builder name, strips unsupported configuration options and then runs default `@angular-devkit/build-angular:extract-i18n` builder. 

For the reference, `nx` is doing the same: https://github.com/nrwl/nx/pull/21802

```
[ ] Yes
[x] No
```


just-jeb:
Hey, thanks for submitting the PR! One question - wouldn't removing plugins mean that the build flow and the extract flow might be different and therefore the extraction might not work as expected?

kamilchlebek:
That's very good question! I'll dig deeper to find out if this may or may not be the issue.

just-jeb:
@kamilchlebek thanks for the update. Seems like we're blocked here until Angular CLI implement your request. Probably submitting a PR could speed it up. 

_Side note: they won't expose the plugins functionality directly from Angular CLI builders, this has been their official policy and that's the reason custom builders exist 😄._

just-jeb:
Closing until it's unblocked

Issue #1730, opened 2024-03-28
ENOENT: no such file or directory, realpath


muuvmuuv:
## Describe the Bug

Angular with latest TypeScript semver range supports direct import.

```
An unhandled exception occurred: ENOENT: no such file or directory, realpath '/Users/marvin/Developer/M8FINDER/M8FINDER/.angular/vite-root/app'
See "/private/var/folders/nb/bwd6_n3d3y583d365z6mcb200000gn/T/ng-Q5hK1L/angular-errors.log" for further details.
```

> https://github.com/just-jeb/angular-builders/blob/master/packages/common/src/load-module.ts#L84-L87

Must be changed to:

```ts
return import(modulePath).then((module) => module.default);
```

## Minimal Reproduction

**Note:** Our policy is that issues reported without a reproduction will be closed immediately and then reopened once a reproduction has been provided. Please respect the developers of this project by doing this. We give of our personal time to work on this project and would rather be spending our time fixing or enhancing the library than chasing down badly described or unreproducible issues.
Please delete this note once you have read it.

## Environment

<pre><code>
Libs
- @angular/core version: 17.3.2
- @angular-devkit/build-angular version: 17.3.2
- @angular-builders/custom-esbuild version: 17.1.1

For Tooling issues:
- Bun version: 1.0.35
- Platform:  Mac

Others:

</code></pre>

just-jeb:
Do you know what's the minimal requirement for Angular/TS here @muuvmuuv ?

muuvmuuv:
I already tried to but its hard to say. We would need tests for that to check against lower semver ranges. But since it no longer works on my end between these and the previous version I would say this is the range.

muuvmuuv:
Would it make sense to grab that here after v18 release since TS version is now high enough?

just-jeb:
Absolutely, I'd say if anything it must be part of v18 release (since as part of breaking changes we can require minimal TS version). 
Do you think you'd have time to make a PR with this change into [v18 PR](https://github.com/just-jeb/angular-builders/pulls)?

muuvmuuv:
Yeah ofc. But I don't think I can write tests for that, Jest is kinda picky with mixed import syntax so it wont be ez.. maybe a custom test suite in shell-script with tsc?

muuvmuuv:
[...]

PR #1860, opened 2024-10-29
refactor: Use new angular/build and modern compiler options

muuvmuuv:
Hey, since NG19 is close to the doors and custom-esbuild still does not use the new `@angular/build` (and therefore no Vite directly) I started working on migrating these. 

Just look at the code not the tools I added or removed, I couldn't get yarn running (it didn't added `@angular/build` in node_modules and other errors) or the tests without the modifications. I have also added my changes from #1730. The full codes runs and tests are successful.

The final build-dist was also tested against an enterprise application I work on, build and dev-server run better than before on ng19. Ng18 was not tested because I think it would be a good start to deprecate or remove some things now.

### DIFF
--- a/packages/custom-esbuild/package.json
+++ b/packages/custom-esbuild/package.json
@@ -30,9 +30,10 @@
     "custom"
   ],
   "scripts": {
-    "prebuild": "yarn clean",
-    "build": "yarn prebuild && tsc && ts-node ../../merge-schemes.ts && yarn postbuild",
-    "postbuild": "yarn run e2e",
+    "prebuild": "pnpm clean",
+    "build": "pnpm prebuild && tsc && tsx ../../merge-schemes.ts",
+    "postbuild": "pnpm run e2e",
+    "watch": "tsc --watch",
     "e2e": "jest --config ../../jest-e2e.config.js",
     "clean": "rimraf dist",
     "ci": "./scripts/ci.sh"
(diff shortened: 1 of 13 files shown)


muuvmuuv:
Just forked your project and modified it a bit to my needs just that you are aware of: https://github.com/muuvmuuv/angular-extended-builders. It will mainly focus on my end to test beta releases on Angular.

just-jeb:
Amazing, thanks for the effort 🙏🏿. It's a great timing for this change, so that we can release it as part of the major version. 
Also thanks for the heads up about your fork. Of course if you implement in your repo anything that you think can be good for everyone you're welcome to back port it here.

muuvmuuv:
Yeah ofc! Once you implement parts of it I will also back port stuff that I add on my end. 

Index transformer is now working as before so compatibility is given to previous versions: https://github.com/muuvmuuv/angular-extended-builders/blob/main/projects/angular-extended-builder/src/application/index.ts#L38

packages/custom-esbuild/package.json
```diff
@@ -30,29 +30,34 @@
...
     "clean": "rimraf dist",
     "ci": "./scripts/ci.sh"
   },
   "builders": "builders.json",
   "dependencies": {
     "@angular-builders/common": "workspace:*",
-    "@angular-devkit/architect": ">=0.1800.0 < 0.1900.0",
-    "@angular-devkit/build-angular": "^18.0.0",
-    "@angular-devkit/core": "^18.0.0"
+		"@angular-devkit/architect": "0.1900.0-next.13",
```
arturovt, line 44: Could you have a look at the indentation? Something is wrong over here...

packages/custom-esbuild/src/dev-server/index.ts
```diff
@@ -42,7 +39,7 @@ export function executeCustomDevServerBuilder(
       const middleware = await Promise.all(
         (options.middlewares || []).map(middlewarePath =>
           // https://github.com/angular/angular-cli/pull/26212/files#diff-a99020cbdb97d20b2bc686bcb64b31942107d56db06fd880171b0a86f7859e6eR52
-          loadModule<Connect.NextHandleFunction>(
+          loadModule<any>(
```
arturovt, line 42: You can take the following file from my PR:

https://github.com/just-jeb/angular-builders/pull/1871/files#diff-5768dc9114dc59a71cc5955282b7b5e1a3e4028ea49a9a753cb66477bb7e59fa

muuvmuuv:
Just got something nice working here feel free to copy it: https://github.com/muuvmuuv/angular-extended-builders/blob/main/projects/angular-extended-builder/schemes.mjs

It generetes correct angular.json Schema (and imo works better than Angular's version)

arturovt (commented):
Can you go through all JSON files, it seems indentation is broken in dependencies.

muuvmuuv:
Oh yeah ofc. I think I had Biome format on safe on and globally tabs activated. Will run the repo formatter

muuvmuuv:
@arturovt I have formatted only all JSON files but with the added prettier script it would also format other TS files as well maybe something worth considering to introduce standard coding styles but I let that decide you

hamfastgamgee:
Now that Angular 19.2 is out and pretty much all of the built-in builders have @angular/build equivalents, @angular-builders/custom-esbuild is the only thing holding my apps from being able to drop all the webpack packages (it's over 300 in my tree) that @angular-devkit/build-angular brings along as dependencies.  Is there any appetite for getting this change done so those of us who have migrated to esbuild can drop a huge number of packages we're no longer using? :)

muuvmuuv:
Guess most is already implemented now

PR #1871, opened 2024-11-21
chore(deps)!: update to Angular 19

arturovt:


just-jeb (approved):
LGMT, one concern, should we merge #1860 into v19 update (as a breaking change)? 
We can also merge the current PR as is but keep it in beta until we merge the breaking changes. 
@arturovt wdyt?

mshima:
Hi, thanks to your work.

Unless a new v18 release is planned I think the v19 migration should be merged before.
- those breaking changes seems to be only related to custom-esbuild which have limited usage.
- this PR is much broader than the custom-esbuild only changes.
- this PR is ready and will allow betas or even a final version of packages except esbuild like happened with jest in the past.

packages/jest/package.json
```diff
@@ -43,16 +43,16 @@
   },
   "dependencies": {
     "@angular-builders/common": "workspace:*",
-    "@angular-devkit/architect": ">=0.1800.0 < 0.1900.0",
-    "@angular-devkit/core": "^18.0.0",
-    "jest-preset-angular": "14.2.4",
+    "@angular-devkit/architect": ">=0.1900.0 < 0.2000.0",
+    "@angular-devkit/core": "^19.0.0",
+    "jest-preset-angular": "14.4.0-rc.0",
```
qmonmert, line 48:
```suggestion
    "jest-preset-angular": "14.4.1",
```

qmonmert (approved):

just-jeb:
@arturovt I agree with @mshima, especially taking into account that #1860 may cause breaking changes.

LanderBeeuwsaert:
It would be nice to get this merged. This is the only package blocking us (and I can imagine that there are other projects in the same situation) from moving to angular 19.

p1n5u:
Is there any timeline when this will be merged? we are in the same position as @LanderBeeuwsaert and this is the only package that is blocking the migration to angular 19.

arturovt:
> @arturovt I agree with @mshima, especially taking into account that #1860 may cause breaking changes.

Ok, lemme rebase and let it go

PR #2194, opened 2026-04-28
fix(common): make ts-node an optional peer dependency (fixes #1638)

### LINKED ISSUE #1638: Make ts-node as optional package
`ts-node` only needed if configuration is written in Typescript. `ts-node` is ugly implementation of TS loader, so most of users go to TSX or manually compile config before

Describe `ts-node` as optional peer dependency ("peerDependencies" + "peerDependenciesMeta")

just-jeb:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`ts-node` is a hard `dependency` of `@angular-builders/common`, so it is always installed even for users whose build configs are plain JavaScript or MJS files that never require TypeScript compilation.

Issue Number: #1638

## What is the new behavior?

`ts-node` is moved to `peerDependencies` with `optional: true` across `@angular-builders/common`, `@angular-builders/custom-esbuild`, `@angular-builders/custom-webpack`, and `@angular-builders/jest`. Users who load TypeScript config files must install `ts-node` explicitly; users with JS/MJS-only configs no longer need it.

```
[x] Yes
[ ] No
```

Users who relied on `ts-node` being auto-installed as a transitive dependency will see a missing-module error at runtime when loading a `.ts` config file.

**Migration path:** If you use TypeScript configuration files (e.g. `webpack.config.ts`, custom esbuild plugins in `.ts`), add `ts-node` explicitly to your project's `devDependencies`:

```
npm install --save-dev ts-node
# or
yarn add --dev ts-node
```

No change is needed if your build config is already JavaScript or MJS.

## Other information

`ts-node` is only invoked inside `loadModule()` in `@angular-builders/common` when the resolved file extension is `.ts`. The runtime guard ensures the optional peer is only required when actually needed.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -25,11 +25,18 @@
   },
   "dependencies": {
     "@angular-devkit/core": "^21.0.0",
-    "ts-node": "^10.0.0",
     "tsconfig-paths": "^4.2.0"
   },
   "devDependencies": {
     "rimraf": "^6.0.0",
     "typescript": "5.9.3"
+  },
+  "peerDependencies": {
+    "ts-node": "^10.0.0"
+  },
+  "peerDependenciesMeta": {
+    "ts-node": {
+      "optional": true
+    }
   }
 }
--- a/packages/custom-esbuild/package.json
+++ b/packages/custom-esbuild/package.json
@@ -47,6 +47,7 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^21.0.0",
     "rxjs": ">=7.0.0",
+    "ts-node": "^10.0.0",
     "vitest": ">=2"
   },
   "devDependencies": {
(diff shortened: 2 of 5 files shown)


just-jeb:
What does it solve? What issue is related to this? Why is this needed?

just-jeb:
This addresses issue #1638: users who write their Angular config files in JavaScript (e.g., `extra-webpack.config.js`, not TypeScript) don't need ts-node installed at all, since ts-node is only used when loading TypeScript config files. However, ts-node is currently a required `dependency` of `@angular-builders/common`, which forces it to be installed for all users even when they don't use TypeScript configs.

Making it an optional peer dependency:
- Reduces install overhead for JS-only config users  
- Makes the dependency relationship more transparent to users
- Still works seamlessly when ts-node is present (for TS configs)

just-jeb:
Closing — #1638 has no community follow-up and ts-node's install footprint isn't a real pain. The PR also has dishonest body claims (no runtime check is being added; README updates are listed but absent). Closing #1638 too. Reopen if signal grows.

PR #2212, opened 2026-05-07
fix(jest)!: scope coverage output per-project in multi-project workspaces (fixes #1009)

### LINKED ISSUE #1009: Support for coverage in multi-project workspaces
Currently if you run `ng test` in a multi-project Angular workspace with coverage enabled, coverage reports are generated with same same filename, and as a result only last project's coverage report "survives".

`jest-junit` reporter is able to generate random report names if `JEST_JUNIT_UNIQUE_OUTPUT_NAME=true` is set, and that seems to be a nice way of handling the situation (at least in my case -- GitLab can just grab and process all files according to a pattern like `junit*.xml`).

**Describe alternatives you've considered**
Currently I'm getting away with the following hack in `jest.config.js`:

```js
module.exports = {
  coverageReporters: [
    [
      'cobertura',
      {
        index: 0,
        get file() {
          return `cobertura-coverage-${ this.index++ }.xml`;
        },
      },
    ],
  ],
};
```

It works fine (and can be easily upgraded to UUID-based filename generation, like `jest-junit` does), but it does look like a hack, and there is no guarantee it will always work because `@angular-builders/jest` could change at some point, e.g. `file` option could stack in some cache or something.

just-jeb:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In multi-project workspaces, all projects write Jest coverage output to the same workspace-root `coverage/` directory. When projects run in sequence (`ng test project-a --coverage; ng test project-b --coverage`), the second run overwrites the first. There is no way to get coverage for multiple projects simultaneously.

Issue Number: #1009

## What is the new behavior?

`coverageDirectory` is set unconditionally in the per-project default config to `<projectRoot>/coverage`. For a standard `ng new` single-project workspace, `projectRoot === workspaceRoot`, so output stays at `./coverage` — no change. For multi-project workspaces, each project writes to its own `projects/<name>/coverage` directory.

Users can still override via their own `jest.config.ts` (the per-project user layer is merged last).

```
[x] Yes
[ ] No
```

For library-only / non-standard workspaces where `projectRoot !== workspaceRoot`, coverage output relocates from `./coverage` to `<projectRoot>/coverage`. Any tooling reading a hardcoded `./coverage/` path is affected. Standard single-project `ng new` workspaces are unaffected. Because this is an observable behavior change, it is labeled `breaking-change` and held for the next major aligned with Angular 22 (not yet released). Will be called out in the CHANGELOG.

## Other information

N/A


just-jeb:
Potentially breaking?

just-jeb:
Re: 'Potentially breaking?'

Yes — it changes coverage output from `<workspace-root>/coverage/` to `<project-root>/coverage/` per project. Intentional fix for the overwrite collision in multi-project workspaces, but for library-only / non-standard workspaces (where `projectRoot !== workspaceRoot`) it relocates the output path, which can break pipelines reading a hardcoded `./coverage/`.

Treating this as a breaking change: labeling `breaking-change` and holding for the next major aligned with Angular 22 (not yet released).

PR #2213, opened 2026-05-07
fix(jest): suppress warning when jest config is provided inline in angular.json (fixes #1102)

### LINKED ISSUE #1102: Q: Why `warning: unable to locate custom jest configuration file`?
https://github.com/just-jeb/angular-builders/blob/aa088a500f40cd802a24840264d6e58e866aa489/packages/jest/src/custom-config.resolver.ts#L24

I was wondering why it's warning me that I don't have a **jest.config.js** in my project. Isn't it just as good to put the config inside **angular.json** or even have no config?

just-jeb:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Projects that don't set `config` in `angular.json` rely on the schema default `jest.config.js`. When that file isn't present (because the user keeps their config in `package.json`, at workspace level, or as an inline JestConfig object), the builder logs `warning: unable to locate custom jest configuration file at path …` on every run. That's the default-discovery path — absence is normal, not user error. See #1102 (`melroy89` 2025-05-25 confirms it still fires on package.json-embedded configs).

Issue Number: #1102

## What is the new behavior?

The warning fires only when the configured path is not the schema default. `config: "jest.config.js"` (the default) absent → silent. `config: "my-custom-jest.config.ts"` absent → still warns, because the user explicitly pointed at a file.

Inline-object and inline-JSON-string configs are unchanged — they already short-circuit before the file-existence check (master, since #7bfe3123).

New unit test: `should NOT warn when the schema default jest.config.js is absent`.

```
[ ] Yes
[x] No
```

The warning text was advisory only; suppressing it for the default-discovery case doesn't change config resolution. Users who relied on the warning to detect a typo'd path still get it.

## Other information

Branch was force-pushed to drop unrelated noise (commitlint downgrade, CHANGELOG entry removals, yarn.lock churn).

### DIFF
--- a/examples/jest/simple-app/angular.json
+++ b/examples/jest/simple-app/angular.json
@@ -106,6 +106,13 @@
             "zoneless": false
           }
         },
+        "test-inline-config": {
+          "builder": "@angular-builders/jest:run",
+          "options": {
+            "config": { "testTimeout": 10000, "verbose": true },
+            "zoneless": false
+          }
+        },
         "e2e": {
           "builder": "@cypress/schematic:cypress",
           "options": {
--- a/examples/jest/simple-app/package.json
+++ b/examples/jest/simple-app/package.json
@@ -7,6 +7,7 @@
     "test": "ng test",
     "test:ts": "ng run simple-app:test-ts-config",
     "test:esm": "ng run simple-app:test-esm-config",
+    "test:inline-config": "ng run simple-app:test-inline-config",
     "lint": "ng lint",
     "e2e": "ng e2e",
     "cypress:open": "cypress open",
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -62,5 +62,5 @@
     "arrowParens": "avoid",
     "singleQuote": true
   },
-  "packageManager": "yarn@3.8.7"
+  "packageManager": "yarn@4.14.1+sha512.64df448055b2d37ba269d7db535a469b8da93f8ef1140c25fd7a83c00a8fbaacb214ca0e02553b92a2c54cef78bb67d0b4817fab02001df0e24fac0faccc3b42"
```
just-jeb: Unclear why this change is here. Unrelated.

just-jeb (commented):
Need to double check that it actually fixes the related issue and addresses the comments.

PR #2309, opened 2026-06-15
test: cypress cache snowball validation (DO NOT MERGE)

just-jeb:
Throwaway validation branch for the Cypress cache fix (#2305). Downgrades Cypress 15.17.0 -> 15.16.0 against a warm cache that holds 15.17.0, to prove the binary served matches the lockfile and does not snowball. Will be closed without merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,6 @@ jobs:
           path: |
             node_modules
             .yarn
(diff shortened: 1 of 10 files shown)


just-jeb:
Validation complete — the dedicated Cypress cache correctly downloads and serves the lockfile's version on a warm cache holding a different version (snowball fixed). Closing without merging.

PR #2311, opened 2026-06-15
test: yarn-dl Cypress exclusion validation (DO NOT MERGE)

just-jeb:
Validates #2310. Bumps Cypress to force a deps-cache miss so the yarn-dl (`~/.cache`) step runs. Checks that its save excludes ~/.cache/Cypress and e2e stays green. Will be closed without merging.


just-jeb:
Validation complete: with path scoped to ~/.cache/node, the yarn-dl cache archives 0 Cypress files (~1MB, corepack only) instead of the ~214MB binary, the dedicated cypress cache still delivers the binary to integration, and the full e2e matrix is green. Closing without merging.

Issue #2420, opened 2026-09-18
custom-esbuild:unit-test: `--coverage` produces no report (empty `coverage*` array options forwarded to `@angular/build`)


myabc:
🤖

## Describe the Bug

Running `ng test --coverage` with `@angular-builders/custom-esbuild:unit-test` collects coverage (`Coverage enabled with v8` is printed) but no report is printed and nothing is written to `coverage/<project>/`. Switching the same target to `@angular/build:unit-test` produces the expected text table and HTML/clover/JSON reports.

Root cause: Angular CLI only skips its "fill undefined arrays with `[]`" schema transform for builders whose name starts with `@angular/build:` ([`architect-base-command-module.ts#L54`](https://github.com/angular/angular-cli/blob/8c438891c9c8be4e66a82148908343befbb462a2/packages/angular/cli/src/command-builder/architect-base-command-module.ts#L54), `isAngularBuild` → [`addUndefinedObjectDefaults`](https://github.com/angular/angular-cli/blob/8c438891c9c8be4e66a82148908343befbb462a2/packages/angular/cli/src/command-builder/architect-base-command-module.ts#L164-L168)). For `@angular-builders/custom-esbuild:unit-test` the full `addUndefinedDefaults` transform runs, so every unset array option arrives as `[]` and is forwarded as-is to `executeUnitTestBuilder`.

`@angular/build` treats those empty arrays as "set by the user" (truthiness checks in [`generateCoverageOption`](https://github.com/angular/angular-cli/blob/8c438891c9c8be4e66a82148908343befbb462a2/packages/angular/build/src/builders/unit-test/runners/vitest/plugins.ts#L515)):

- `coverageReporters: []` → `coverage.reporter: []`, which overrides Vitest's defaults and any `coverage.reporter` from the runner config. No reporter runs, so nothing is printed or written.
- `coverageInclude: []` → `coverage.include: ['spec-*.js', 'chunk-*.js']`. With `excludeAfterRemap: true`, the remapped source paths (`src/**/*.ts`) never match, so every file is filtered out. Even with reporters passed explicitly the report is `0/0`.
- `coverageExclude: []` takes the "user provided excludes" branch needlessly (harmless, but also not what the official builder sees).

The wrapper already works around exactly this problem for one option:

https://github.com/just-jeb/angular-builders/blob/4c61b61d5e0f29c01f641832f2ea52e42bcecea8/packages/custom-esbuild/src/unit-test/index.ts#L17-L19

[...]

PR #2422, opened 2026-09-21
fix(custom-esbuild): drop empty coverage arrays in unit-test builder

myabc:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

With `@angular-builders/custom-esbuild:unit-test`, `ng test --coverage` collects coverage (`Coverage enabled with v8` is printed) but prints no coverage table and writes nothing to `coverage/<project>/`.

Angular CLI only skips its "fill unset array options with `[]`" schema transform for builders named `@angular/build:*`. For this builder the unset `coverageReporters`, `coverageInclude` and `coverageExclude` options arrive as `[]` and are forwarded unchanged. `@angular/build` treats an empty array as a value the user set: `coverageReporters: []` replaces Vitest's default reporters, so none run, and `coverageInclude: []` becomes `['spec-*.js', 'chunk-*.js']`, which filters out every source file once coverage is remapped to the original sources.

Issue Number: #2420

Reproduction: https://github.com/myabc/custom-esbuild-coverage-repro (`HEAD~1` is the control case with the stock `@angular/build:unit-test` builder).

## What is the new behavior?

The unit-test builder already deleted an empty `browsers` array for the same reason. That guard now also covers `coverageInclude`, `coverageExclude` and `coverageReporters`, so the options reach `executeUnitTestBuilder` as `undefined`, exactly as they do for `@angular/build:unit-test`. `ng test --coverage` then runs Vitest's default reporters and writes the report to `coverage/<project>/`. Non-empty arrays are passed through unchanged.

Tests: new `src/unit-test/index.spec.ts` covers the four stripped options, the pass-through of non-empty arrays, and the forced `vitest` runner. Docs: a pitfall row in `packages/custom-esbuild/AGENTS.md`; the README needs no change because no option or usage changes.

```
[ ] Yes
[x] No
```

## Other information

Verified by hand: the equivalent change applied to `dist/unit-test/index.js` of `22.0.1` makes `ng test --coverage` write a report in the reproduction project and in the project where I first hit this. Locally I ran only the new spec; I did not run the full build or the integration tests, so I am relying on CI for those.

Other unset array options (`setupFiles`, `exclude`, `reporters`, …) presumably arrive as `[]` too. I only traced the coverage ones and kept the list explicit; happy to switch to stripping every empty array generically if you prefer.


just-jeb:
Thanks for the bug report and the fix. 
One question, what happens if user indeed sets these options to empty array explicitly?

myabc:
> One question, what happens if user indeed sets these options to empty array explicitly?

@just-jeb the short answer is that an explicit `[]` gets dropped as well – i.e. it's the same as unset.

I think this is OK since the CLI also cannot tell them apart for us. `addUndefinedDefaults` turns unset into `[]` for any non-`@angular/build:*` builder, so both cases reach the builder as `[]`. Only `@angular/build:unit-test` itself gets the `addUndefinedObjectDefaults` path that keeps unset as `undefined`.

myabc:
While tracing the other options I found `reporters: []` is a second live instance of the same bug (Angular's Vitest plugin deletes `testConfig.reporters` whenever the option is defined), so I switched to stripping every empty array in 2472d77 and retitled the PR.

just-jeb:
Thanks, the analysis of what `@angular/build` does with `[]` checks out.

About not being able to tell them apart, `context.getTargetOptions(context.target)` gives you the raw angular.json options (+ configurations) before the schema transform runs, so a key the user explicitly wrote is there.

It matters cuz without it this is a breaking change. Someone relying on `include: []` to run nothing gets a patch update and suddenly the whole suite runs, which can easily break their build (that's Copilot's comment above).

So can we only drop the empty arrays that aren't in the raw options? Plus a spec case for an explicit `[]` surviving, and an integration case in `tests/integration.js` running `--coverage` that fails if `coverage/` is empty. The spec mocks `executeUnitTestBuilder`, so it never sees the CLI filling in `[]`.

Only gap I see is `[]` coming from a CLI flag (not sure that's even possible). Wdyt?

